PR #25, opened 2025-01-07
docs/vectorizer readme

AllenWriter:
- **Fix:The two parameters given by the vectorizer.ai official API are API Id and API Secret. The original hint was misleading.**
- **Docs:updat vectorizer readm**


Yeuoly (approved):
LGTM

PR #146, opened 2025-01-14
fix deepseek chat misalign

hjlarry:
fix https://github.com/langgenius/dify/issues/12621

### DIFF
--- a/models/deepseek/manifest.yaml
+++ b/models/deepseek/manifest.yaml
@@ -15,7 +15,7 @@ meta:
     entrypoint: main
     language: python
     version: '3.12'
-  version: 0.0.2
+  version: 0.0.1
 name: deepseek
 plugins:
   models:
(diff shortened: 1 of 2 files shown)


Yeuoly (changes requested):
Please also bump the version of deepseek to 0.0.2

models/deepseek/manifest.yaml
```diff
@@ -15,7 +15,7 @@ meta:
     entrypoint: main
     language: python
     version: '3.12'
-  version: 0.0.1
+  version: 0.0.2
```
Yeuoly: please keep meta.version in 0.0.1, it's the manifest standard version,

Yeuoly (changes requested):

models/deepseek/manifest.yaml
hjlarry: ok

Yeuoly (approved):
LGTM

PR #180, opened 2025-01-23
add models for siliconflow

senseb:
add models for siliconflow


Yeuoly:
Hi, single PR should contains changes of only one plugin, please split them to 2 PR instead, also, please bump the versions

PR #189, opened 2025-02-08
feat(siliconflow): add the latest models in siliconflow (include r1)

Nov1c444:
add the latest models in siliconflow (include r1)
image


Yeuoly (approved):
LGTM

PR #202, opened 2025-02-11
Hy/chore with dify main

yihong0618:
core with main repo

### DIFF
--- a/models/openai/manifest.yaml
+++ b/models/openai/manifest.yaml
@@ -21,7 +21,7 @@ plugins:
   models:
     - "provider/openai.yaml"
 meta:
-  version: 0.0.1
+  version: 0.0.2
   arch:
     - "amd64"
     - "arm64"
--- a/models/openai/models/llm/llm.py
+++ b/models/openai/models/llm/llm.py
@@ -1290,34 +1290,37 @@ class OpenAILargeLanguageModel(_CommonOpenAI, LargeLanguageModel):
 
             # calculate num tokens for function object
             num_tokens += len(encoding.encode("name"))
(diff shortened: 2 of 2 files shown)


Yeuoly:
Hi, single PR can only contributes to one plugin, and it should be split to servals smalls

yihong0618:
> Hi, single PR can only contributes to one plugin, and it should be split to servals smalls

Hi the change almost the same, more can check https://github.com/langgenius/dify/pull/13518

Yeuoly:
> > Hi, single PR can only contributes to one plugin, and it should be split to servals smalls
> 
> Hi the change almost the same, more can check [langgenius/dify#13518](https://github.com/langgenius/dify/pull/13518)

I understand, but as you can see, after PR has been merged, a plugin will been packaged and uploaded to marketplace, we are planning to support multiple plugin changes in the same PR, but once an error occurs, just partial plugins succeed, It's hard to rollback changes, that's why only one plugin is supported.

Yeuoly:
BTW, please also bump the version in each manifest.yaml (not meta.version), it's helpful for version management.

yihong0618:
> > > Hi, single PR can only contributes to one plugin, and it should be split to servals smalls
> 
> > 
> 
> > Hi the change almost the same, more can check [langgenius/dify#13518](https://github.com/langgenius/dify/pull/13518)
> 
> 
> 
> I understand, but as you can see, after PR has been merged, a plugin will been packaged and uploaded to marketplace, we are planning to support multiple plugin changes in the same PR, but once an error occurs, just partial plugins succeed, It's hard to rollback changes, that's why only one plugin is supported.

ok will do this tomorrow

PR #242, opened 2025-02-23
set service_account_info as None when vertex_service_account_key param is not set

shohei-ihaya:
When I leave blank `Service Account Key` field to use ADC(Application Default Credential), got error "vertex_service_account_key"
image

To fix this issue, set service_account_info as None when vertex_service_account_key param is not set.


crazywoola:

shohei-ihaya:
@crazywoola updated version, and sync main.

crazywoola (approved):

PR #245, opened 2025-02-23
update llms from siliconflow

senseb:


senseb:
Update llm changes from siliconflow.

PR #289, opened 2025-02-27
refactor: use override rather than write a new function

acelyc111:
Related to https://github.com/langgenius/dify/pull/14449


crazywoola:
Please bump the version in the manifest as well :)

Issue #358, opened 2025-03-03
bug(gemini): All models are not working var 0.0.7 by an error that displays description.


Rion-Sato:
Error Content:
PluginInvokeError: {"args":null,"error_type":"PluginDaemonInnerError","message":"encountered an error: invalid character '\u003c' looking for beginning of value status: 503 Service Temporarily Unavailable original response: \u003chtml\u003e"}

note:Same error in all models

rion-x-coffee:
After a few hours, I was able to solve the problem!(not merged PR #359)

PR #359, opened 2025-03-03
fix(gemini): Bug Fix(Downgrade dify_plugin library the same as var 0.0.6)

Rion-Sato:
fix langgenius/dify-official-plugins#358


crazywoola (approved):

crazywoola:
Hello, there are several issues with our cloud services, I think this is not the root cause of this problem.

rion-x-coffee:
After a few hours, I was able to solve the problem!

PR #365, opened 2025-03-03
chore: gpustack tools api path

alexcodelf:
This PR is going to:
1. Change GPUStack tools API path from `v1-openai` to `v1`
2. Add timeouts for image generation tools, as some LLM-based image generators can take a long time to process requests, potentially causing excessive wait times for users.

### DIFF
--- a/tools/gpustack/manifest.yaml
+++ b/tools/gpustack/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.1
+version: 0.0.2
 type: plugin
 author: langgenius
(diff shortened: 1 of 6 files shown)


crazywoola:

alexcodelf:
updated.

crazywoola:
Emm, I think we can use this https://github.com/langgenius/dify-official-plugins/pull/382 instead. Since it includes all the changes here.

crazywoola (approved):

PR #372, opened 2025-03-04
fix: missing import `time` module

chekun:
Line 326 use time.sleep, but it was never been imported.

It should be well tested before it came to v1.0.0


laipz8200 (approved):
LGTM. Thank you for your contribution!

PR #376, opened 2025-03-04
change author name

lichengwu:
change author name


crazywoola:
A quick reminder, the name should be written in English and lowercased. Also, please update the version number in manifest as well.

crazywoola:
If you want to keep the name of yours, you need to submit to this repo instead.

PR #377, opened 2025-03-04
fix(tongyi): Update stream response handling in Tongyi LLM model

Snuffel233:
- Modify `incremental_output` parameter to match stream mode
- Add `is_reasoning` flag to track reasoning state during response generation
- Refactor response handling to improve stream processing logic


crazywoola:
Please bump the version as well.

PR #406, opened 2025-03-09
feat(azure_openai): add o1 support for azure_openai

dengquan:


kurokobo:
Please test it before sending PR.
I don't believe this works as expected.

kurokobo:
#409

PR #434, opened 2025-03-11
Tavily search and extract images to files[array] | Email html convert and add attachments 

mendybraun:


mendybraun:
1. **Tavily search and extract:** map image URLS from the search and extract to the files[array] variable. (so it can be used in downstream nodes - like the LLM for vision).
2. **Email:** add email option to convert markdown content to HTML and added option to add attachments to the email

crazywoola:
> 1. **Tavily search and extract:** map image URLS from the search and extract to the files[array] variable. (so it can be used in downstream nodes - like the LLM for vision).
> 2. **Email:** add email option to convert markdown content to HTML and added option to add attachments to the email

```
PR files cross different plugin directories, skip upload.
```

Please split the PR into 2 PRs, because each one should only contain one tool or model runtime due to CI restriction.

mendybraun:
@crazywoola thanks for your comment
I added https://github.com/langgenius/dify-official-plugins/pull/437 for tavily changes
and https://github.com/langgenius/dify-official-plugins/pull/439 for email changes

PR #437, opened 2025-03-12
Tavily search and extract images to files[array]

mendybraun:
Tavily search and extract: map image URLS from the search and extract to the files[array] variable. (so it can be used in downstream nodes - like the LLM for vision).

### DIFF
--- a/tools/tavily/manifest.yaml
+++ b/tools/tavily/manifest.yaml
@@ -32,4 +32,4 @@ resource:
 tags:
 - search
 type: plugin
-version: 0.0.6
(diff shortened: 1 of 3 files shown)


tools/tavily/manifest.yaml
```diff
@@ -32,4 +32,4 @@ resource:
 tags:
 - search
 type: plugin
-version: 0.0.3
+version: 0.0.6
```
crazywoola:
```suggestion
version: 0.0.4
```

crazywoola (approved):

PR #462, opened 2025-03-14
Supports setting the model's UI display name for openai_api_compatible.

bxfxf:
Close https://github.com/langgenius/dify/issues/14122


models/openai_api_compatible/provider/openai_api_compatible.yaml
```diff
@@ -23,6 +23,15 @@ model_credential_schema:
       en_US: Enter full model name
       zh_Hans: 输入模型全称
   credential_form_schemas:
+    - variable: show_name
```
crazywoola, line 26: I would suggest using `display_name`.

bxfxf: > I would suggest using `display_name`.

OK

PR #475, opened 2025-03-17
feat: support 'encouraged_response_prefix' parameter of LLM

acelyc111:
https://github.com/langgenius/dify/pull/15945


crazywoola:
Here are few things you might need to know:
- CI only checks one tool/model at one time, if a PR includes multiple models or tools, the check will be failed, you should split them into 2 PRs.
- Please bump the version for the upgrade.

PR #485, opened 2025-03-18
fix: tongyi stream function call bug

cyflhn:


jqhr:
图片
This code concat function call parameters ,  before tool call, the agent need to parse json type parameters, it went wrong, like .
```python
Run failed: Expecting property name enclosed in double quotes: line 1 column 19 (char 18)
```
The error above is in fc_agent_runner.py line 333.
图片

Now I cannot find any solutions to figure out it

jqhr:
Oh, I find the latest version figure out this error.
Great!

PR #544, opened 2025-03-22
fix: space missing when format

alterxyz:
update manifest


tools/telegraph/requirements.txt
```diff
@@ -1,2 +1,2 @@
 dify_plugin==0.0.1b65
```
crazywoola: dify_plugin==0.0.1b74

Yeuoly (approved):

Issue #524, opened 2025-03-21
Failed to transform agent message with Tongyi v0.0.13


canyangyang:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [Dify issues](https://github.com/langgenius/dify/issues) & [Dify Official Plugins](https://github.com/langgenius/dify-official-plugins/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.0.1

### Plugin version

0.0.13

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Looks like another json parse problem
It`s fixed in tongyi  v0.0.12 ,but come back again when updated to v0.0.13……
Please have a check locally before push to the origin

![Image](https://github.com/user-attachments/assets/81c08248-fca2-44e7-9a19-3b1713e9faf3)

### ✔️ Error log

Failed to transform agent message: PluginlnvokeError: {"args":{},"error_tyKeyError","me
ssage":" current_timecurrent_time'"}

zddwild:
I have a similar issue. When I choose the tongyi v0.0.14 plugin to use in the agent, and invoke the tool using the faction calling mode, an exception is always thrown, which seems to be related to JSON parsing. However, this situation does not occur with the deepseek plugin. Even though I use the function calling mode in the same way, it still works fine, although it’s a bit slower，my dify version is 1.1.2
<img width="835" alt="Image" src="https://github.com/user-attachments/assets/b9f6ec37-1140-4523-9bfc-dc77b60a9ba2" />

wangdisdu:
this pr can fix it  https://github.com/langgenius/dify-official-plugins/pull/557

PR #557, opened 2025-03-25
fix: tongyi llm stream function call incremental output duplidate bug

wangdisdu:
tongyi function call is not support incremental output

### DIFF
--- a/models/tongyi/models/llm/llm.py
+++ b/models/tongyi/models/llm/llm.py
@@ -247,7 +247,7 @@ class TongyiLargeLanguageModel(LargeLanguageModel):
         )
         return result
 
-    def _handle_tool_call_stream(self, response, tool_calls):
+    def _handle_tool_call_stream(self, response, tool_calls, incremental_output):
         tool_calls_stream = response.output.choices[0].message["tool_calls"]
         for tool_call_stream in tool_calls_stream:
             idx = tool_call_stream.get('index')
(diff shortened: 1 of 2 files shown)


wangdisdu:
this pr can fix issue 524 
https://github.com/langgenius/dify-official-plugins/issues/524

youyuanrsq:
I'm also experiencing this problem and now I can only fall back to an older version to make it usable

models/tongyi/models/llm/llm.py
```diff
@@ -335,7 +341,7 @@ def _handle_generate_stream_response(
                 )
                 if not resp_content:
                     if "tool_calls" in response.output.choices[0].message:
-                        self._handle_tool_call_stream(response, tool_calls)
+                        self._handle_tool_call_stream(response, tool_calls, False)
```
youyuanrsq, line 344: The incremental_output should be True

wangdisdu:
merge太慢了。
我看到，这个pr改动的所有代码在4月3号的这个commit里已经被提交的主分支上了：63164a1430bb1f20b018faf71c6375bab2f61ae7

models/tongyi/models/llm/llm.py
wangdisdu: 我不清楚为什么这里需要为True。
另外这个commit把本pr的所有代码提交到主分支了：63164a1430bb1f20b018faf71c6375bab2f61ae7

PR #624, opened 2025-03-31
Fix the issue of concurrent calls causing the 'think' tag to become d…

alex-sample-code:
Fix: concurrent calls causing the 'think' tag to become disordered


crazywoola:
Please fill in the pull request template. Eg. https://github.com/langgenius/dify-official-plugins/pull/631

PR #1329, opened 2025-07-23
feat(gemini):Improvements for Gemini 2.5 Series(Thinking mode)

liiir1985:
## Related Issues or Context
-Fixed the token usage count calculation for gemini, prompt token usage count was counted multiple times in streaming mode
-Add CoT support for Gemini 2.5 series, and added the thinking budget setting to control it.

- [x] My Changes Affect Token Consumption Metrics

Fixed the token calculation, which was wrong.
Before:
The prompt token usage count was ridiculously high.
image

After:
It shows the actual token usage count
image

- [x] My Changes Affect Other LLM Functionalities (Reasoning Process, Grounding, Prompt Caching, etc.)

Added CoT support for Gemini 2.5 series, can use the thinking_budget to control the token used for CoT and to enable/disable the thinking mode.
The content of CoT is surrounded with ```<thinking></thinking>```.
image
image

- [X] I have Bumped Up the Version in Manifest.yaml 

## Dify Plugin SDK Version
- [x] I have Ensured `dify_plugin>=0.3.0,<0.5.0` is in requirements.txt ([SDK docs](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/README.md))

- [x] Dify Version is: 1.6.0, I have Tested My Changes on Local Deployment Dify with a Clean Environment That Matches the Production Configuration. 

### SaaS Environment

### DIFF
--- a/models/gemini/models/llm/gemini-2.5-flash-lite-preview-06-17.yaml
+++ b/models/gemini/models/llm/gemini-2.5-flash-lite-preview-06-17.yaml
@@ -34,8 +34,8 @@ parameter_rules:
     required: false
   - name: thinking_budget
     default: 8192
-    min: 128
-    max: 32768
+    min: 512
+    max: 24576
     label:
       en_US: Thinking Budget
       zh_Hans: 思考预算
--- a/models/gemini/models/llm/gemini-2.5-flash-lite.yaml
+++ b/models/gemini/models/llm/gemini-2.5-flash-lite.yaml
@@ -32,6 +32,17 @@ parameter_rules:
       zh_Hans: 仅从每个后续标记的前 K 个选项中采样。
       en_US: Only sample from the top K options for each subsequent token.
     required: false
+  - name: thinking_budget
+    default: 8192
+    min: 512
+    max: 24576
+    label:
+      en_US: Thinking Budget
+      zh_Hans: 思考预算
+    type: int
+    help:
+      en_US: The token budget used for CoT of the model, if you don't need CoT then leave this parameter off
+      zh_Hans: 大模型用于思维链的Token预算，如果不需要思维链则保持该开关关闭
   - name: max_output_tokens
     use_template: max_tokens
     default: 65536
--- a/models/gemini/models/llm/gemini-2.5-flash-preview-04-17.yaml
+++ b/models/gemini/models/llm/gemini-2.5-flash-preview-04-17.yaml
@@ -34,7 +34,7 @@ parameter_rules:
   - name: thinking_budget
     default: 8192
     min: 128
-    max: 32768
+    max: 24576
     label:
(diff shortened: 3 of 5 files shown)


juniorsereno:
Nice!

QIN2DIM:
@liiir1985 There may be bugs with changes to the thinking budget that result in invocation behavior that is not as expected.

```python

class ThinkingConfig(_common.BaseModel):
  """The thinking features configuration."""

  include_thoughts: Optional[bool] = Field(
      default=None,
      description="""Indicates whether to include thoughts in the response. If true, thoughts are returned only if the model supports thought and thoughts are available.
      """,
  )
  thinking_budget: Optional[int] = Field(
      default=None,
      description="""Indicates the thinking budget in tokens. 0 is DISABLED. -1 is AUTOMATIC. The default values and allowed ranges are model dependent.
      """,
  )
```

jortegac:
@liiir1985 Would you be able to implement the same for the vertex_ai plugin? 

https://github.com/langgenius/dify-official-plugins/tree/main/models/vertex_ai

liiir1985:
> @liiir1985 There may be bugs with changes to the thinking budget that result in invocation behavior that is not as expected.
> 
> ```python
> class ThinkingConfig(_common.BaseModel):
>   """The thinking features configuration."""
> 
>   include_thoughts: Optional[bool] = Field(
>       default=None,
>       description="""Indicates whether to include thoughts in the response. If true, thoughts are returned only if the model supports thought and thoughts are available.
>       """,
>   )
>   thinking_budget: Optional[int] = Field(
>       default=None,
>       description="""Indicates the thinking budget in tokens. 0 is DISABLED. -1 is AUTOMATIC. The default values and allowed ranges are model dependent.
>       """,
>   )
> ```
@QIN2DIM 
You are right, current implementation doesn't support automatic thinking budget. It can only enable/disable thinking mode, and setting a fixed thinking_budget. The reason for this is because the behaviours between pro/flash/flash-lite are not consistent, and I'm uncomfortable with comparing the model name to some hard coded strings, since the model name is changing frequently. Of course we can also leave the decision to the users, they'll be responsible if the llm node throws exception once they set the parameters incorrectly. I'm not sure which way is better, any ideas?

liiir1985:
> @liiir1985 Would you be able to implement the same for the vertex_ai plugin?
> 
> https://github.com/langgenius/dify-official-plugins/tree/main/models/vertex_ai

I'm sorry, I'm afraid I'm not able to do so, because I don't have access to vertex_ai personally so I don't have the necessary environment to do the testing

QIN2DIM:
@liiir1985 I am editing a new version to try to solve this problem in a best practice way. The google-genai interface changes frequently, which is very annoying. Half a month ago, thinking_budget was still a number that could not be less than 0.

PR #1695, opened 2025-09-11
feat: add gemini_image (available for nano banana)

Zhaohuai-L:
## Related Issues or Context
This PR adds support for google's image generation and editing models including gemini-2.5-flash-image-preview ( also known as Nano banana)
This PR is based on demo code of gemini, refering to https://github.com/GoogleCloudPlatform/generative-ai/blob/main/gemini/getting-started/intro_gemini_2_5_image_gen.ipynb

Image Generation:

Image Editing:

## This PR contains Changes to *Non-Plugin* 

- [ ] Documentation
- [ ] Other

## This PR contains Changes to *Non-LLM Models Plugin*
- [x] I have Run Comprehensive Tests Relevant to My Changes

- [ ] My Changes Affect Message Flow Handling (System Messages and User→Assistant Turn-Taking)

- [ ] My Changes Affect Tool Interaction Flow (Multi-Round Usage and Output Handling, for both Agent App and Agent Node)

- [ ] My Changes Affect Multimodal Input Handling (Images, PDFs, Audio, Video, etc.)

- [ ] My Changes Affect Multimodal Output Generation (Images, Audio, Video, etc.)

- [ ] My Changes Affect Structured Output Format (JSON, XML, etc.)

- [ ] My Changes Affect Token Consumption Metrics

- [ ] My Changes Affect Other LLM Functionalities (Reasoning Process, Grounding, Prompt Caching, etc.)

- [ ] Other Changes (Add New Models, Fix Model Parameters etc.)

- [ ] I have Bumped Up the Version in Manifest.yaml (Top-Level `Version` Field, Not in Meta Section)

## Dify Plugin SDK Version
- [ ] I have Ensured `dify_plugin>=0.3.0,<0.5.0` is in requirements.txt ([SDK docs](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/README.md))

- [ ] Dify Version is: , I have Tested My Changes on Local Deployment Dify with a Clean Environment That Matches the Production Configuration. 

### SaaS Environment

### DIFF
--- a/tools/gemini_image/README.md
+++ b/tools/gemini_image/README.md
@@ -1,0 +1,40 @@
+# Gemini Image
+
+## Overview
+
+The **Gemini Image Tool** is used to access Gemini's image generation and editing models on Vertex AI.
+
+---
+
+## Configuration
+
+### 1. Get a Google Cloud API key
+
+1. Go to the [Google Cloud ](https://cloud.google.com/vertex-ai/generative-ai/docs/start/api-keys?usertype=existinguser).  
+2. Follow the steps to create an API key.  
+
+
+### 3. Gemini Image Tool in Dify
+
+1. In the **Dify Console**, go to **Plugin Marketplace**.  
+2. Search for **Gemini Image** and install it.
+
+### 4. Configure in Dify
+
+In **Dify Console > Tools > Gemini Image > Authorize**, enter:  
+
+- **Vertex Service Account key**: The key from [Google Cloud](https://ai.google.dev/gemini-api/docs/api-key) in base64 format.  
+- **project ID**: The [Google Cloud project ID](https://cloud.google.com/resource-manager/docs/creating-managing-projects#identifying_projects).  
+- **vertex_location**: The [Google model endpoint location](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations).
+---
+
+## Usage
+
+The Gemini Image Tool can be used in the following application types:
+
+### Chatflow / Workflow Applications
+Add a **Gemini Image node** to generate or edit images during flow execution.
+
+### Agent Applications
+Enable the **Gemini Image tool** in Agent applications.  
+


tools/gemini_image/README.md
crazywoola, line 1: Please do not leave this blank

Zhaohuai-L: @crazywoola done, please check.

crazywoola (approved):

yt-koike:
Hello, @Zhaohuai-L !

Thanks a lot for bringing Nano Banana to Dify — I really appreciate it!  
I think this is a great step toward enabling image outputs from Gemini.  
I have three small suggestions that might help improve the usability and flexibility of your plugin in the future:

## 1. Use Web API instead of Python package  
As you may know, Google also provides a [REST API for Gemini Image](https://ai.google.dev/gemini-api/docs/image-generation#rest).  
I think using this would make the code simpler, easier to understand, and easier to modify.  
It’s also convenient because the only credential you need is an API Key (a string like `"AIza..."`).  
You can get your own API Key [here](https://aistudio.google.com/app/apikey).  
Unlike `service_account.Credentials.from_service_account_info()`, you don’t need the full JSON data.

## 2. Add more details to README.md  
Right now, README.md is missing some key information.  
For example, Step 4 (which should probably be Step 3) says:  
“**Vertex Service Account key**: The key from [Google Cloud](https://ai.google.dev/gemini-api/docs/api-key) in base64 format.”  
But it doesn’t explain what this *key* actually looks like.  

Looking at your code, it seems to expect JSON data since it goes into `service_account.Credentials.from_service_account_info()`, which takes [a dict](https://google-auth.readthedocs.io/en/latest/_modules/google/oauth2/service_account.html#Credentials.from_service_account_info).  
But the term *API Key* usually makes me think of a single string with random characters.  

I assume you’ve just been too busy to update it, so no worries at all.  
You don’t have to fix it yourself — you could simply open an issue to point it out, and others can help improve it.

## 3. Stop using Base64
I’m also curious why the API Key has to be in Base64 format. Did it not work when using the JSON directly?  
It’d be helpful if you could share the reason, so we can discuss whether Base64 is actually needed for credential handling.

Many thanks once again. I truly look forward to the future of this wonderful plugin!

yt-koike:
And one more little thing here, you should not make a generation request for validating credentials because it costs the user's quota.
Even if making the request is the only way to validate it, you should only do minimal validations in _validate_credentials() like checking if the URL is alive.

PR #2458, opened 2026-01-20
fix: 404 on nested groups (>2 levels) and missing slash in raw file URL

joeyhacker:
Problem
When repository is filled with a 3-level group path like Bigdata/vcrm/SparkStreaming, the plugin URL-encodes the whole string → Bigdata%2Fvcrm%2FSparkStreaming; old GitLab instances return 404 for such "deep" encoded paths.
When fetching single file content, the raw-url is built as …/repository/files{path}/raw (no slash), causing another 404.
Root cause
urllib.parse.quote(identifier, safe="") encodes all slashes.
String concatenation forgets a separator / before {encoded_item_path}.
Solution
Replace full quote with identifier.replace("/", "%2F") to keep only two levels.
Add missing / in raw-file URL template.
Result
Nested-group repos can now be traversed and single files downloaded without 404.
No breaking change for existing 2-level or numeric-id usages.
Closes #<issue-number>

### DIFF
--- a/tools/gitlab/manifest.yaml
+++ b/tools/gitlab/manifest.yaml
@@ -30,4 +30,4 @@ resource:
       enabled: true
 tags: []
 type: plugin
-version: 0.0.6
+version: 0.0.7


joeyhacker:
@austinbcdev  need review the PR

crazywoola:
Please bump the version in the mainifest as well.

PR #2627, opened 2026-02-27
feat(vertex_ai): add gemini-3.1-flash-image-preview model

GOSICK-Angel:


WH-2099:
Thank you for your contribution, but I noticed that your PR does not have a corresponding Issue and the PR description does not use a template. This does not conform to the standardized development process we are promoting. Please add the corresponding Issue and use the template in both the Issue and the PR for verification.

To allow for better follow-up, this PR will be closed, but you are very welcome to open a new PR.

PR #2908, opened 2026-04-20
fix(tongyi/tts): multi-sentence streaming is cut off after first sentence

Zenine:
## Problem

When the user-provided `content_text` exceeds the model's `word_limit`,
`_tts_invoke_streaming` splits it into several sentences and loops
`SpeechSynthesizer.call()` in a background thread. The consumer generator
only yields audio from the **first** sentence — all remaining sentences
are synthesised in the background but their audio is discarded. End users
hear a truncated clip; wall-clock and API-cost is spent on output nobody
receives.

## Root cause

Each `SpeechSynthesizer.call()` is synchronous: it returns only after
`on_complete` / `on_close` / `on_error` fires. The current `Callback`
enqueues the `None` end-of-stream sentinel in each of those, so:

1. Sentence 1 synthesises → `on_complete` puts `None` → `.call()` returns.
2. Consumer loop sees `None` → `break` → generator ends.
3. Worker thread happily starts sentence 2, but nobody is reading the queue.

## Fix

Move the end-of-stream sentinel to the producer thread's `finally` block
so it is emitted exactly once, after **all** sentences have been
synthesised (or after an exception). `on_error` still enqueues `None`
so errors terminate the stream immediately, as before.

No API change. Behaviour for short text (single sentence, no split) is
unchanged because the thread's `finally` still runs after the single
`.call()` completes.

## Test plan

- [x] Short text (below `word_limit`) — one sentence, unchanged behaviour.
- [x] Long text (above `word_limit`, splits into N sentences) — all N
      sentences' audio now reaches the consumer; previously only the
      first sentence's audio was yielded.
- [x] `on_error` path still short-circuits (verified by asserting the
      consumer raises `InvokeBadRequestError` when the dashscope
      callback simulates an error response).

Diff is +27/−21 lines, all inside `_tts_invoke_streaming` and `Callback`.

### DIFF
--- a/models/tongyi/manifest.yaml
+++ b/models/tongyi/manifest.yaml
@@ -25,5 +25,5 @@ resource:
     model:
       enabled: false
 type: plugin
-version: 0.1.36
+version: 0.1.37
 created_at: "2024-12-10T16:13:50.29298939+08:00"
--- a/models/tongyi/models/tts/tts.py
+++ b/models/tongyi/models/tts/tts.py
@@ -68,7 +68,8 @@ class TongyiText2SpeechModel(_CommonTongyi, TTSModel):
         ws_base_address = get_ws_base_address(credentials)
         try:
             audio_queue: Queue = Queue()
-            callback = Callback(queue=audio_queue)
+            failure_event = threading.Event()
+            callback = Callback(queue=audio_queue, failure_event=failure_event)
 
             def invoke_remote(content, v, api_key, cb, at, wl, base_address):
                 try:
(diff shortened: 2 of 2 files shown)


Zenine:
Thanks for the review! All three suggestions are addressed in `8e4e94ed`:

1. **Early-stop on failure** — added a shared `threading.Event` (`failure_event`). When `on_error` fires, the producer checks it at the top of each iteration and breaks out of the loop, so the remaining sentences don't consume API quota after the consumer has already torn down.
2. **Telemetry headers** — `SpeechSynthesizer.call` in the streaming path now passes `headers=BURY_POINT_HEADER`, matching the existing `_process_sentence` helper.
3. **Redundant sentinels** — `Callback.on_error` no longer enqueues `None`; the producer's `finally` block is now the single source of truth for the end-of-stream sentinel, so the queue can never contain duplicates.

Also bumped `models/tongyi/manifest.yaml` to `0.1.37` — `0.1.36` is already on the marketplace, which was what the previous Pre PR Check was failing on. CI is green now.

Issue #2881, opened 2026-04-14
Azure OpenAI Image Plugin API method bug when using OpenAI Image 1.5


jcamkh:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [Dify issues](https://github.com/langgenius/dify/issues) & [Dify Official Plugins](https://github.com/langgenius/dify-official-plugins/issues), including closed ones.
- [x] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

v1.13

### Plugin version

v0.0.3

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

Hi Dify team,

Hope you are doing well. I would like to report an issue for the Azure OpenAI Image Plugin https://marketplace.dify.ai/plugin/langgenius/azure_openai_tool

After updating from the old gpt-image-1 to image-1.5, the image generation function works normally but not when in editing mode 
![Image](https://github.com/user-attachments/assets/e6c48059-dc62-4497-b1f4-999825a90785)

It showed the follwoing error message:
"Unsupported content type: 'multipart/form-data'. This API method only accepts 'application/json'"

Perhaps it has to do with a change of API format, could you kindly help to take a look at it? Many thanks

### ✔️ Error log

"{\n  \"files\": [],\n  \"text\": \"Failed to edit image: Error code: 400 - {'error': {'message': \\\"Unsupported content type: 'multipart/form-data'. This API method only accepts 'application/json' requests, but you specified the header 'Content-Type: multipart/form-data'. Please try again with a supported content type.\\\", 'type': 'invalid_request_error', 'param': None, 'code': 'unsupported_content_type'}}\"\n}"

nikkoyuthinkcol:
@crazywoola can the Dify team look at this issue? Thanks.

kurokobo:
@jcamkh @nikkoyuthinkcol 
I was going to look into this, but I couldn't reproduce the issue on my end.
If possible, could you share a minimal, reproducible DSL file?

Also, could you let me know which API version you set with your credential for the tool?

jcamkh:
[...]

PR #2995, opened 2026-04-28
feat(azure_openai_tool): add gpt-image-2

leslie2046:
## Related Issues or Context

图片
图片

## This PR contains Changes to *Non-Plugin* 

- [ ] Documentation
- [ ] Other

## This PR contains Changes to *Non-LLM Models Plugin*
- [x] I have Run Comprehensive Tests Relevant to My Changes

- [ ] My Changes Affect Message Flow Handling (System Messages and User→Assistant Turn-Taking)

- [ ] My Changes Affect Tool Interaction Flow (Multi-Round Usage and Output Handling, for both Agent App and Agent Node)

- [ ] My Changes Affect Multimodal Input Handling (Images, PDFs, Audio, Video, etc.)

- [ ] My Changes Affect Multimodal Output Generation (Images, Audio, Video, etc.)

- [ ] My Changes Affect Structured Output Format (JSON, XML, etc.)

- [ ] My Changes Affect Token Consumption Metrics

- [ ] My Changes Affect Other LLM Functionalities (Reasoning Process, Grounding, Prompt Caching, etc.)

- [ ] Other Changes (Add New Models, Fix Model Parameters etc.)

- [x] I have Bumped Up the Version in Manifest.yaml (Top-Level `Version` Field, Not in Meta Section)

## Dify Plugin SDK Version
- [x] I have Ensured `dify_plugin>=0.3.0,<0.6.0` is declared in `pyproject.toml` and locked in `uv.lock`, or kept in `requirements.txt` only for legacy plugins without `uv.lock` ([SDK docs](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/README.md))

- [x] Dify Version is: , I have Tested My Changes on Local Deployment Dify with a Clean Environment That Matches the Production Configuration. 

### SaaS Environment


kurokobo:
Thanks for working on this, but AFAIK if we specify the deployment name in azure_openai_api_model_name, we can use gpt-image-2 even with plugins that have already been released as version 0.0.6.
See the screenshots in my previous PR: https://github.com/langgenius/dify-official-plugins/pull/2946

So, could you please clarify the strong reason for making such a large-scale modification?

leslie2046:
> Thanks for working on this, but AFAIK if we specify the deployment name in azure_openai_api_model_name, we can use gpt-image-2 even with plugins that have already been released as version 0.0.6. See the screenshots in my previous PR: #2946
> 
> So, could you please clarify the strong reason for making such a large-scale modification?

Thank you for your reminder. I also found that gpt-image-2 can be used without any modifications, but the resolution cannot be set arbitrarily because the resolution parameter of the previous tool was of type select.

Do you have a better way to accommodate the limitations of older models and the needs of newer models?

kurokobo:
@leslie2046 
Oh, I hadn’t realized that there were different parameters available. Thanks for pointing that out!
Given that, I can understand implementing it as a separate tool. It would be nice if we could refactor the shared parts for better reusability, but I think it’s fine to leave that for later :)

One thing I wanted to mention is that including `model` in the API request can cause errors in environments where the deployment name has been changed from the default, so I removed it in my PR.

- Context: https://github.com/langgenius/dify-official-plugins/issues/2881#issuecomment-4296395856
- Fix: https://github.com/langgenius/dify-official-plugins/pull/2946/changes#diff-470252748a9897246abf41d0e14e512e2de6dc7e37a72579368a08943826e032

I haven’t been able to test your code, but since your current PR includes `model` as a parameter for both `generate` and `edit`, it might potentially cause the same issue. If everything works fine without it, maybe it’s better to remove it.

Anyway thanks for working on this :)

leslie2046:
@kurokobo 
Thank you for pointing that out. I also think we could simply change the parameter type to a regular string, based on the previous version. What do you think?

kurokobo:
@leslie2046 
Hmm, I think changing the type of an existing parameter could break backward compatibility.
If we want to minimize changes to the existing code, one possible approach might be something like this:

- In addition to the existing `size` parameter, add a string-type `custom_size` parameter.
- Clearly state that this is a parameter for gpt-image-2 based deployment, and that if it is set, it takes precedence and `size` is ignored.
- When constructing parameters for the API, if `custom_size` is present, use it instead of `size`.

crazywoola (approved):

PR #3201, opened 2026-05-26
feat(bedrock): attach Dify app_id as Converse requestMetadata (opt-in)

mas-sakai:
## Summary

Add opt-in support for forwarding the Dify `app_id` to Amazon Bedrock's Converse API `requestMetadata` field so CloudWatch invocation logs can be filtered per Dify app for cost and usage tracking.

- New provider credential `enable_request_metadata` (select, default `disabled`) exposed in `provider/bedrock.yaml`.
- When enabled, `_generate_with_converse` reads `app_id` from the current Dify session via `dify_plugin.get_current_session()` and attaches `{dify_app_id, dify_source}` to `requestMetadata` on Converse / ConverseStream calls.
- New helper module `models/llm/_metadata.py` with `normalize_metadata_value` (sanitizes to the Bedrock-allowed character set `[a-zA-Z0-9\s:_@$#=/+,-.]`, truncates to 256 chars, preserves mixed case) and `build_dify_request_metadata` (returns `None` for empty `app_id` so the caller can skip attaching).
- Scope: Converse route only. The legacy `InvokeModel` path does not support `requestMetadata` and is intentionally untouched.
- Graceful degradation: `dify_plugin` is imported lazily inside a `try/except ImportError`, so SDK versions without `get_current_session` (currently anything before sdks#313 ships) just skip metadata attachment without raising.
- Version bump `0.0.66` → `0.0.67`.

## Why opt-in

Default `disabled` to preserve current behavior for tenants who don't want Dify identifiers landing in their CloudWatch logs. Operators opt in per-credential in the Dify console.

## Temporary SDK pin

`pyproject.toml` points `dify_plugin` at the `feat/pass-session-to-model-plugins` branch of `ryuta-kobayashi-ug/dify-plugin-sdks` (PR langgenius/dify-plugin-sdks#311 / commit reference #313) because `get_current_session()` is not yet available in a released SDK. The branch SDK is `0.9.0` and requires `tiktoken>=0.12.0`, so the bedrock `tiktoken` constraint is bumped accordingly and `uv.lock` is refreshed.

**Action item before merge:** once the SDK PR is merged and a new release is published, revert this branch pin back to a normal release-version constraint. Calling out explicitly so it isn't shipped as-is.

## Files

- `models/bedrock/provider/bedrock.yaml` — new `enable_request_metadata` credential entry.
- `models/bedrock/models/llm/_metadata.py` — new helpers + tests.
[...]

PR #3203, opened 2026-05-26
feat(openai): attach Dify app_id as request metadata (opt-in)

mas-sakai:
## Summary

Adds opt-in support for attaching the Dify `app_id` as OpenAI request metadata, so usage on the OpenAI Usage Dashboard can be filtered per Dify app.

When the new `enable_request_metadata` credential is set to `enabled`, the plugin reads `app_id` from the current Dify session (via `get_current_session()` from the SDK) and attaches `{dify_app_id, dify_source}` as the `metadata` field on both `chat.completions.create` and `responses.create`. The default is `disabled`, so behavior is unchanged unless the operator opts in.

This is the OpenAI counterpart of the same feature already shipped for Vertex AI (#3168) and Bedrock (#3201). The responsibility split mirrors those plugins:

| Plugin   | What the plugin attaches              | What the operator must enable      |
| -------- | ------------------------------------- | ---------------------------------- |
| Vertex AI| labels on `generateContent`           | BigQuery billing export            |
| Bedrock  | `requestMetadata` on Converse         | CloudWatch invocation logging      |
| OpenAI   | `metadata` on Chat / Responses        | Stored Completions on the OpenAI account — *the plugin enables `store=true` automatically* |

### Design note: `store` is set to `true` alongside metadata

E2E verification against a live OpenAI account showed the API **rejects `metadata` when `store` is false**:

```
BadRequestError: The 'metadata' parameter is only allowed when 'store' is enabled.
```

Enabling the metadata feature therefore inherently requires `store=true`. The plugin makes this explicit: `apply_dify_metadata_if_enabled` sets `store=true` alongside the metadata (a non-destructive merge that respects an explicit `store` value already on the request). This means requests and responses are persisted to **Stored Completions** on the OpenAI account when the operator opts in. The credential's help text in `provider/openai.yaml` documents this storage behavior in both English and Simplified Chinese, and links to the OpenAI `store` reference.

Bedrock (#3201) and Vertex AI (#3168) do not have this constraint, so their plugins remain unchanged.

Related:
- langgenius/dify#35772
- langgenius/dify-plugin-sdks#311
- langgenius/dify-plugin-sdks#313 (SDK branch this plugin currently pins)
- #3168 (Vertex AI counterpart)
- #3201 (Bedrock counterpart)

> [!NOTE]
> **Draft / dependency.** `pyproject.toml` temporarily pins `dify_plugin` to the fork branch `ryuta-kobayashi-ug/dify-plugin-sdks@feat/pass-session-to-model-plugins` because it carries the `get_current_session()` API used here (langgenius/dify-plugin-sdks#313). The pin will revert to a versioned `dify_plugin` spec once that SDK change merges and a release ships.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| _Usage Dashboard rows are anonymous; per-app filtering is impossible._ | _With opt-in enabled, the plugin sets `store=true` and attaches `dify_app_id` and `dify_source`, so requests are recorded to Stored Completions and can be filtered per app._ |

End-to-end screenshots will be added once the dependent SDK change (#313) ships and a Dify deployment can be exercised against a live OpenAI account with Stored Completions enabled.

## LLM Plugin Checklist

<details>
<summary>Areas affected by this change</summary>

- [x] Other LLM functionality (reasoning, grounding, prompt caching, etc.) — request metadata for billing/observability

</details>

The change is additive and behind a credential that defaults to `disabled`. No existing message flow, tool, multimodal, structured-output, or token-accounting code path is altered. Only `_chat_generate` and `_build_responses_api_params` gain a one-line opt-in hook that mutates the request kwargs immediately before the OpenAI client call.

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (`0.4.1` → `0.4.3`)
- [x] `dify_plugin` declared in `pyproject.toml` and locked in `uv.lock` (currently pinned to the SDK branch carrying #313 — see Draft note above)

## Testing

- [x] Unit tests — `uv run pytest tests/` (22 passed): covers normalization (UUID passthrough, punctuation, mixed case, non-ASCII, 512-char truncation, empty string, non-string coercion), `build_dify_metadata` (None / empty / source marker / UUID passthrough / length normalization), and `apply_dify_metadata_if_enabled` (no-op on missing or `disabled` credential, silent on session-lookup failure, non-destructive metadata merge, and the new `store=true` behavior: sets `store=true` on opt-in, preserves an explicit caller-supplied `store`, and never touches `store` when disabled).
- [ ] Local deployment — Dify version: pending (blocked on SDK #313 release; will retest after the pin is reverted)
- [ ] SaaS (cloud.dify.ai)

### DIFF
--- a/models/openai/manifest.yaml
+++ b/models/openai/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.4.1
+version: 0.4.2
 type: plugin
 author: "langgenius"
 name: "openai"
--- a/models/openai/manifest.yaml
+++ b/models/openai/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.4.2
+version: 0.4.3
 type: plugin
 author: "langgenius"
 name: "openai"
--- a/models/openai/models/llm/_metadata.py
+++ b/models/openai/models/llm/_metadata.py
@@ -13,12 +13,15 @@ Unlike Bedrock or Vertex, OpenAI does not document a character pattern
 restriction on metadata values, so ``normalize_metadata_value`` only
 enforces string coercion and the 512-character length cap.
 
-This plugin deliberately does NOT set ``store=true``. Whether stored
-completions are retained — and therefore whether this metadata appears
-in the dashboard — is governed by the account-level Stored Completions
-setting, which the terminus owner controls. Setting ``store`` from the
-plugin would change persistence behavior as a side effect of an
-observability opt-in, which is out of scope for this feature.
+The OpenAI API only accepts ``metadata`` when ``store`` is true — it
+rejects the request with ``BadRequestError: The 'metadata' parameter is
+only allowed when 'store' is enabled.`` otherwise. Enabling this feature
+therefore inherently requires ``store=true``, so
+``apply_dify_metadata_if_enabled`` sets it alongside the metadata. This
+means requests and responses are persisted to Stored Completions on the
+OpenAI account; the credential's help text documents that storage
+behavior. An explicit ``store`` value already present on the request is
+respected rather than overwritten.
 """
 
 from __future__ import annotations
@@ -67,8 +70,11 @@ def apply_dify_metadata_if_enabled(target: dict, credentials: dict) -> None:
     """Inject Dify metadata into ``target`` when opt-in credential is set.
 
     Reads ``credentials['enable_request_metadata']``; when ``"enabled"``,
-    resolves the current Dify session's ``app_id`` (best-effort) and sets
-    ``target['metadata']`` to the built dict, if one is produced.
+    resolves the current Dify session's ``app_id`` (best-effort), sets
+    ``target['metadata']`` to the built dict (if one is produced), and sets
+    ``target['store'] = True`` — the API only accepts ``metadata`` when
+    ``store`` is enabled. An explicit ``store`` value already on ``target``
+    is left untouched.
 
     Session lookup failures are swallowed silently: metadata attachment is
     telemetry, and must never break generation if the SDK is missing or
@@ -99,3 +105,7 @@ def apply_dify_metadata_if_enabled(target: dict, credentials: dict) -> None:
         target["metadata"] = {**existing, **metadata}
     else:
         target["metadata"] = metadata
+    # The API only accepts metadata when store=true. Don't overwrite an
+    # explicit store value already set by the caller.
+    if "store" not in target:
+        target["store"] = True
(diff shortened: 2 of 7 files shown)


ryuta-kobayashi-15:
Superseded by #3688.

Two things make this branch unrecoverable as-is.

It pins `dify_plugin` to `git+https://github.com/ryuta-kobayashi-ug/dify-plugin-sdks.git@feat/pass-session-to-model-plugins`. That account no longer exists, so the fork is gone and the dependency can't resolve. The pin was a stopgap while the SDK change was unreleased; `get_current_session()` is now public in `dify_plugin` v0.10.0.

More significantly, #3419 rewrote this plugin for 1.0 and split the `models/llm/llm.py` this patch edits into `chat.py` / `responses.py` / `stream.py` / `tokens.py`. The functions the patch hooks into, `_build_responses_api_params` and the `extra_model_kwargs` assembly, no longer exist, so this is a port rather than a rebase.

#3688 opens from a fresh branch on current `main` with:

- the metadata call placed in `chat.py::generate_chat`, right after `_add_identity`, where credentials are already in scope
- `responses.py::parameters()` taking an optional `credentials` argument, so the existing four-positional-argument call sites keep working unchanged, with `stream.py` threaded through
- encrypted reasoning preserved. The API only accepts `metadata` when `store` is true, and `main` only requests `reasoning.encrypted_content` when `store` is false. Applied naively, enabling this credential would silently change multi-turn reasoning on gpt-5 / o-series models. The port tracks whether `store=true` came from telemetry alone and keeps requesting the encrypted payload in that case
- metadata skipped when the caller set `store=false` explicitly, which is a guaranteed 400
- version bumped 1.0.3 → 1.0.4 across `manifest.yaml`, `pyproject.toml` and the hardcoded assertion in `tests/test_catalog.py`
- README updated for the new credential, and the "Reasoning state" section corrected since it stated `store=false` unconditionally

The feature-disabled path is unchanged: `parameters()` produces identical output for `credentials` of `None`, `{}` and `disabled`.

I opened a new PR instead of updating this one because I don't have push access to this fork's branch.

@mas-sakai could you close this one when you get a chance? I don't have permission to close it myself. Happy to fold anything back in if you'd rather keep this PR as the one of record.

Mairuis:
Closing this superseded draft as noted by the author. The replacement PR #3688 has already been merged. Please continue any follow-up work from the current main branch.

PR #3233, opened 2026-05-31
feat(azure_openai): attach Dify app_id as request metadata (opt-in)

mas-sakai:
## Summary

Adds opt-in support for attaching the Dify `app_id` as Azure OpenAI request metadata, so per-app usage can be filtered in Stored Completions.

When the new `enable_request_metadata` credential is set to `enabled`, the plugin reads `app_id` from the current Dify session (via `get_current_session()` from the SDK) and attaches `{dify_app_id, dify_source}` as the `metadata` field on both the Chat Completions route (`chat.completions.create`) and the Responses route (`responses.create`). The default is `disabled`, so behavior is unchanged unless the operator opts in.

This is the Azure OpenAI counterpart of the same feature already shipped for Vertex AI (#3168), Bedrock (#3201), and OpenAI (#3203). The responsibility split mirrors those plugins:

| Plugin       | What the plugin attaches              | What the operator must enable             |
| ------------ | ------------------------------------- | ----------------------------------------- |
| Vertex AI    | labels on `generateContent`           | BigQuery billing export                   |
| Bedrock      | `requestMetadata` on Converse         | CloudWatch invocation logging             |
| OpenAI       | `metadata` on Chat / Responses        | Stored Completions on the OpenAI account — *the plugin enables `store=true` automatically* |
| Azure OpenAI | `metadata` on Chat / Responses        | Stored Completions on the Azure OpenAI resource — *the plugin enables `store=true` automatically* |

### Design note: `store` is set to `true` alongside metadata

E2E verification against a live Azure OpenAI resource showed the API **rejects `metadata` when `store` is false**:

```
BadRequestError: The 'metadata' parameter is only allowed when 'store' is enabled.
```

Enabling the metadata feature therefore inherently requires `store=true`. The plugin makes this explicit: `apply_dify_metadata_if_enabled` sets `store=true` alongside the metadata (a non-destructive merge that respects an explicit `store` value already on the request). This means requests and responses are persisted to **Stored Completions** on the Azure OpenAI resource when the operator opts in — the same data-storage context as Azure OpenAI's standard logging. Per Microsoft's data privacy documentation, this data stays within the customer's Azure environment: it does not leave Azure and is not used to train foundation models. The credential's help text in `provider/azure_openai.yaml` documents this storage behavior in both English and Simplified Chinese.

Bedrock (#3201) and Vertex AI (#3168) do not have this constraint, so their plugins remain unchanged.

See:
- https://learn.microsoft.com/en-us/azure/foundry/responsible-ai/openai/data-privacy
- https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/stored-completions

Related:
- langgenius/dify#35772
- langgenius/dify-plugin-sdks#311
- langgenius/dify-plugin-sdks#313 (SDK branch this plugin currently pins)
- #3168 (Vertex AI counterpart)
- #3201 (Bedrock counterpart)
- #3203 (OpenAI counterpart)

> [!NOTE]
> **Draft / dependency.** `pyproject.toml` temporarily pins `dify_plugin` to the fork branch `ryuta-kobayashi-ug/dify-plugin-sdks@feat/pass-session-to-model-plugins` because it carries the `get_current_session()` API used here (langgenius/dify-plugin-sdks#313). The pin will revert to a versioned `dify_plugin` spec once that SDK change merges and a release ships.

## Change Type

- [x] LLM plugin

## End-to-end verification

Verified end-to-end against a live Azure OpenAI resource ahead of the SDK
dependency (#313). The plugin's metadata-building logic was exercised
directly against `chat.completions.create` on a deployment with
`enable_request_metadata=enabled`. Three test requests were persisted to
Stored Completions on the Azure OpenAI resource with the plugin-supplied
`metadata` intact, retrieved via the `GET /openai/v1/chat/completions`
listing API (verification UUID `8ec4083b-8e36-4174-a3a2-291e0aba0ec7`):

```json
{
  "object": "list",
  "data": [
    {
      "id": "chatcmpl-DoSqrw1FaKzr73LAM3RpfqwBqgkgt",
      "model": "gpt-4o-2024-11-20",
      "metadata": {
        "dify_app_id": "8ec4083b-8e36-4174-a3a2-291e0aba0ec7",
        "dify_source": "dify"
      },
      "choices": [{"index": 0, "message": {"content": "2", "role": "assistant"}, "finish_reason": "stop"}],
      "usage": {"total_tokens": 18, "completion_tokens": 2, "prompt_tokens": 16}
    },
    {
      "id": "chatcmpl-DoSqqQrsQfpfjCT57E0m1Qfhqfs3b",
      "metadata": {
        "dify_app_id": "8ec4083b-8e36-4174-a3a2-291e0aba0ec7",
        "dify_source": "dify"
      },
      "choices": [{"index": 0, "message": {"content": "1", "role": "assistant"}, "finish_reason": "stop"}]
    }
    // and one more, all with the same dify_app_id / dify_source
  ]
}
```

This also confirms the design note above: the `metadata` parameter is only
accepted by the Azure OpenAI API when `store=true`, so the plugin sets both
together. The 400 error reproduced without `store=true`:

```
BadRequestError: The 'metadata' parameter is only allowed when 'store' is enabled.
```

The full Dify-integrated path (where #313 supplies `app_id` to the plugin
session) will be re-verified once #313 ships, but the plugin's terminal-side
behavior is confirmed working.

## LLM Plugin Checklist

<details>
<summary>Areas affected by this change</summary>

- [x] Other LLM functionality (reasoning, grounding, prompt caching, etc.) — request metadata for billing/observability

</details>

The change is additive and behind a credential that defaults to `disabled`. No existing message flow, tool, multimodal, structured-output, or token-accounting code path is altered. Only `_chat_generate` (Chat Completions route) and `_chat_generate_with_responses` (Responses route) gain a one-line opt-in hook that mutates the request kwargs immediately before the Azure OpenAI client call. The legacy `_generate` completions route is intentionally left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/models/azure_openai/manifest.yaml
+++ b/models/azure_openai/manifest.yaml
@@ -24,4 +24,4 @@ resource:
     model:
       enabled: false
 type: plugin
-version: 0.0.59
+version: 0.0.60
--- a/models/azure_openai/models/llm/_metadata.py
+++ b/models/azure_openai/models/llm/_metadata.py
@@ -1,0 +1,97 @@
+"""Helpers for attaching Dify metadata to OpenAI chat/responses requests.
+
+OpenAI's ``chat.completions.create`` and ``responses.create`` accept a
+``metadata`` field (a string-to-string map) that — when the account has
+Stored Completions enabled — surfaces in the Usage Dashboard, enabling
+per-app filtering. Constraints, from the OpenAI API reference:
+
+  Set of 16 key-value pairs that can be attached to an object. Keys are
+  strings with a maximum length of 64 characters. Values are strings with
+  a maximum length of 512 characters.
+
+Unlike Bedrock or Vertex, OpenAI does not document a character pattern
+restriction on metadata values, so ``normalize_metadata_value`` only
+enforces string coercion and the 512-character length cap.
+
+This plugin deliberately does NOT set ``store=true``. Whether stored
+completions are retained — and therefore whether this metadata appears
+in the dashboard — is governed by the account-level Stored Completions
+setting, which the terminus owner controls. Setting ``store`` from the
+plugin would change persistence behavior as a side effect of an
+observability opt-in, which is out of scope for this feature.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Optional
+
+_MAX_VALUE_LENGTH = 512
+
+
+def normalize_metadata_value(s: Any) -> str:
+    """Normalize an arbitrary value into an OpenAI metadata value.
+
+    Coerces non-string input via ``str()`` so that, e.g., a numeric ``0``
+    becomes ``"0"`` rather than being silently dropped by the empty-check,
+    then truncates to 512 characters. OpenAI does not document a
+    character-pattern restriction, so no character substitution is
+    performed.
+    """
+    if not isinstance(s, str):
+        s = str(s)
+    if not s:
+        return ""
+    return s[:_MAX_VALUE_LENGTH]
+
+
+def build_dify_metadata(app_id: Any) -> Optional[dict[str, str]]:
+    """Build the Dify metadata dict for an OpenAI request, or return ``None``.
+
+    Returns ``None`` if ``app_id`` is ``None`` or an empty string, so the
+    caller can skip attaching metadata entirely. Other falsy values (e.g.
+    numeric ``0``) are coerced by ``normalize_metadata_value`` and pass
+    through. Otherwise, returns a dict with ``dify_app_id`` (normalized)
+    and a static ``dify_source`` marker.
+    """
+    if app_id is None or app_id == "":
+        return None
+    return {
+        "dify_app_id": normalize_metadata_value(app_id),
+        "dify_source": "dify",
+    }
+
+
+def apply_dify_metadata_if_enabled(target: dict, credentials: dict) -> None:
+    """Inject Dify metadata into ``target`` when opt-in credential is set.
+
+    Reads ``credentials['enable_request_metadata']``; when ``"enabled"``,
+    resolves the current Dify session's ``app_id`` (best-effort) and sets
+    ``target['metadata']`` to the built dict, if one is produced.
+
+    Session lookup failures are swallowed silently: metadata attachment is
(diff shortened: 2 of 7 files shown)


ryuta-kobayashi-15:
Superseded by #3687.

This branch pins `dify_plugin` to `git+https://github.com/ryuta-kobayashi-ug/dify-plugin-sdks.git@feat/pass-session-to-model-plugins`. That account no longer exists, so the fork is gone and the dependency can't resolve — CI here can no longer install, regardless of the code.

That pin was a stopgap while the SDK change was unreleased. It has since shipped: `get_current_session()` is public in `dify_plugin` v0.10.0. `main` has since moved to `dify_plugin>=0.10.2`, which already satisfies it.

#3687 opens from a fresh branch on current `main` with:

- the SDK pin dropped in favour of `main`'s, and `uv.lock` regenerated
- the version bump redone on top of `main` (0.0.66 → 0.0.67, since #3683 had already taken 0.0.66)
- two guards that prevent requests the API rejects outright:
  - metadata is skipped when the effective api-version predates Stored Completions. `openai_api_version` is optional and the fallback default is `2024-02-15-preview`, while the dropdown still offers versions back to `2023-12-01-preview`. On this branch, enabling the credential on a default configuration would have failed **every** generation, and `validate_credentials` can't catch it because it builds its own probe request. Versionless `/openai/v1` endpoints are exempt, matching how `common.py` omits `api_version` for them
  - metadata is skipped when the caller set `store=false` explicitly. This branch attached metadata anyway, which is a guaranteed 400, and `test_metadata.py` asserted that shape as correct

The retention wording in the help text carried over as-is; it was the clearest of the four.

I opened a new PR instead of updating this one because I don't have push access to this fork's branch.

@mas-sakai could you close this one when you get a chance? I don't have permission to close it myself. Happy to fold anything back in if you'd rather keep this PR as the one of record.

Mairuis:
Closing this superseded draft as noted by the author. The replacement PR #3687 has already been merged. Please continue any follow-up work from the current main branch.

PR #3250, opened 2026-06-04
Upgrade Firecrawl plugin to the v2 API

rakshith48:
Upgrades the Firecrawl plugin endpoints `/v1`→`/v2` (matching Dify's own v2 core extractor). Notable v2 changes handled: structured extraction moves from a top-level `extract` field to a `formats: [{type: json, schema, prompt}]` entry; `systemPrompt` (removed from v2's json format) is folded into `prompt`; `ignoreSitemap`→`sitemap` enum. User-facing tool params/outputs unchanged.

**Verification:** static + SDK-introspection/mocked against the real v2 SDK (`firecrawl-py 4.28.2` / `@mendable firecrawl v4`); not run against the live API. Happy to address review/CI feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/firecrawl/manifest.yaml
+++ b/tools/firecrawl/manifest.yaml
@@ -32,4 +32,4 @@ tags:
   - search
   - utilities
 type: plugin
-version: 0.0.9
+version: 0.1.0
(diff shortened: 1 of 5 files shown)


rakshith48:
Good catch on the stale `extract` string — fixed, with a small divergence from the exact suggestion:

The suggested `if "extract" in formats or len(json_format) > 1:` removes `extract` but then appends a bare `{"type": "json"}` in the *extract-without-schema/prompt* case, which v2 also rejects (json extraction needs a schema or prompt). So instead I **always** strip `extract` (never valid in v2) and only append the json format when a schema/prompt is actually provided:

```python
# v2 dropped the "extract" format string entirely — always remove it.
formats = [f for f in formats if f != "extract"]
# Only request json (structured) extraction when a schema/prompt was provided.
if len(json_format) > 1:
    formats.append(json_format)
```

This avoids sending either the invalid `extract` string or a contentless json format.

crazywoola (approved):

PR #3267, opened 2026-06-09
fix ollama model provider category

crazywoola:
## Summary
- remove the bundled Ollama web tool provider from the Ollama model plugin
- keep the plugin manifest in the model-provider category only
- document Ollama web search/fetch as external cloud APIs that should be exposed via a separate tool plugin or search tool

## Why
The marketplace upload for `models/ollama` fails with:

```text
change in category not allowed
```

Ollama is a model provider plugin, so adding `plugins.tools` changes the marketplace category. This follow-up keeps the latest model/embedding updates from #3266 while removing the category-changing tool provider files.

## Validation
- `python -m py_compile models/ollama/models/llm/llm.py models/ollama/models/text_embedding/text_embedding.py models/ollama/tests/test_llm.py`
- `UV_PROJECT_ENVIRONMENT=.ci-test-uv-venv uv run --frozen --with pytest pytest` from `models/ollama`
- manifest check confirms `plugins` contains only `models`
- `git diff --check origin/main...HEAD`


crazywoola:
Addressed the review note in d6d8d312 by removing the unused tool permission from models/ollama/manifest.yaml. The Ollama plugin now declares only model permissions and only plugins.models in the manifest.

PR #3329, opened 2026-06-20
fix(gemini): prevent embedding crashes on text splitting and missing token usage

### LINKED ISSUE #3323: [gemini] embedding crashes — (A) infinite recursion, (B) NoneType in np.average
#### Environment
- Dify: **1.14.2**
- `dify_plugin` SDK in plugin venv: **0.9.0**
- Plugin: **`langgenius/gemini` v0.8.4 and v0.8.1**
- Model: `gemini-embedding-2-preview` (used by a Knowledge Base / knowledge-retrieval node)

Both bugs are in `models/text_embedding/text_embedding.py` and crash any Knowledge Retrieval / embedding call that exercises them, killing the workflow run mid-generation (the run errors before the message is persisted, so the conversation appears lost).

---

#### Error signature
```
PluginInvokeError: [models] Error: maximum recursion depth exceeded
  (dataset_retrieval -> retrieval_service -> embedding_search)
```

#### Root cause
The recursive text splitter uses `context_size` (a **token** count) as a **character** index:
```python
context_size = self._get_context_size(model, credentials)   # e.g. 2048 (TOKENS)
...
if num_tokens >= context_size:
    cutoff = context_size                                     # used as a CHAR offset
    for pattern in [r"[.!?]", r",", r"\s"]:
        match = re.search(pattern, text[context_size:])       # char slicing
        if match:
            cutoff = context_size + match.start() + 1
            break
    splitted_text.extend(self._split_texts_to_fit_model_specs(
        client, model, [text[:cutoff]], context_size))        # HEAD recursion
    splitted_text.extend(self._split_texts_to_fit_model_specs(
        client, model, [text[cutoff:]], context_size))        # TAIL recursion
```
When `num_tokens(text) >= context_size` **but** `len(text) <= context_size` (i.e. a text shorter than `context_size` characters that still counts as `>= context_size` tokens — e.g. CJK, code, base64, or token-dense content), the HEAD call `[text[:cutoff]]` passes the **same text** unchanged (no progress) → infinite recursion → `RecursionError`.

#### Suggested fix
Replace the HEAD recursion with a non-recursive append (the HEAD is at most ~`context_size` chars and fits the model); only the TAIL `[text[cutoff:]]` recurses, and it is strictly shorter than `text`, so it always terminates:
```python
head_tokens = self._count_tokens(client, model, text[:cutoff])
splitted_text.append((text[:cutoff], head_tokens))
```

---

#### Error signature
```
PluginInvokeError: [models] Error: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
```

#### Root cause
In the "merge embeddings by averaging" block:
```python
embeddings, num_tokens = zip(*embeddings)
if len(embeddings) == 1:
    embedding = embeddings[0]
else:
    average = np.average(embeddings, axis=0, weights=num_tokens)   # CRASH
```
When the Gemini API omits token usage in the embedding response, `num_tokens` is a tuple of `None`. NumPy then performs `None + None` while computing the weighted average → `TypeError`. (Note: the `used_tokens` sum just below already handles `None` via `used_token or chunk_size`, but the `np.average` call does not.)

#### Suggested fix
Guard `weights` and fall back to an unweighted average when any token count is `None`:
```python
_w = [t for t in num_tokens]
if any(x is None for x in _w):
    _w = None
average = np.average(embeddings, axis=0, weights=_w)
```

---

Both fixes are minimal, preserve the original behaviour on the happy path, and only change behaviour in the edge cases that currently crash. Happy to open a PR if useful.

Shevilll:
## Summary

Fixes #3323

Two crashes in the Gemini text embedding plugin (`models/gemini/models/text_embedding/text_embedding.py`) that kill any Knowledge Retrieval / embedding call exercising them.

### Bug A — `maximum recursion depth exceeded` in `_split_texts_to_fit_model_specs`
`context_size` is a **token** budget but was used as a **character** index when slicing text. When a text counts as `>= context_size` tokens while being `<= context_size` characters (token-dense content like CJK, code, or base64), the head slice `text[:cutoff]` equaled the unchanged input, so the recursion never made progress → `RecursionError`.

**Fix:** derive the cutoff in characters from the token-to-character ratio and clamp it to `[1, len(text) - 1]`, so both the head and tail are strictly shorter than the input and the recursion always terminates. Punctuation/whitespace boundaries are still preferred but can never empty the tail.

### Bug B — `unsupported operand type(s) for +: 'NoneType' and 'NoneType'` in `np.average`
When the API omits token usage, `num_tokens` is a tuple of `None`, and `np.average(..., weights=num_tokens)` crashes.

**Fix:** fall back to an unweighted average when any weight is missing — consistent with the existing `used_token or chunk_size` handling in the usage summation right below.

Both fixes are minimal and preserve happy-path behavior; they only change the edge cases that currently crash.

## Change Type

## Testing
Added unit tests in `models/gemini/models/tests/test_embedding.py` (mocked, no API key required):
- splitter terminates on token-dense and long inputs, with lossless reassembly and every chunk within `context_size`;
- merge path averages cleanly when per-chunk token counts are `None` (unweighted) and honors weights when present.

Ran the full gemini embedding suite in a clean Python 3.12 venv matching `pyproject.toml`/`uv.lock`: **51 passed**. Also reproduced both original crashes in isolation and confirmed the fixes resolve them.


crazywoola:

crazywoola (approved):

crazywoola (approved):

PR #3330, opened 2026-06-20
fix(mistralai): report real embedding latency instead of constant zero

ly-wang19:
## Summary

`MistralAITextEmbeddingModel._calc_response_usage` reports
`latency=time.time() - time.time()`, which is **always ≈ 0**, so every Mistral
embedding call records zero latency in its `EmbeddingUsage`.

The inline comment ("Will be set by the framework") is a misunderstanding: the
framework sets `self.started_at` (base `AIModel` in
`dify_plugin/interfaces/model/ai_model.py` does `self.started_at = time.perf_counter()`
before `invoke`), but the **provider** is responsible for computing the latency
from it. This aligns Mistral with the SDK's own contract and every sibling
provider:

- SDK base: `dify_plugin/interfaces/model/openai_compatible/text_embedding.py` → `latency=time.perf_counter() - self.started_at`
- Siblings: `openai`, `cohere`, `azure_openai`, `voyage`, … all use `time.perf_counter() - self.started_at`

Fix: `latency=time.perf_counter() - self.started_at`. No behaviour change beyond
correct latency reporting. `models/mistralai` version bumped 0.0.9 → 0.0.10.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

N/A — internal usage-metric fix, no UI/output change.

| Before | After |
| ------ | ----- |
| `usage.latency` always `0.0` | `usage.latency` reflects real elapsed time |

## LLM Plugin Checklist

This is a one-line fix to the **usage-latency metric only** — it does not touch
model invocation, credentials, parameters, or embedding output, so the inference
test matrix does not apply. Verified locally:

- `time.time() - time.time()` is constant `0.0` (5/5 samples); `time.perf_counter() - self.started_at` reflects real elapsed time (e.g. `0.5s`).
- `self.started_at` is guaranteed set by the base `AIModel` before `_invoke` runs (`ai_model.py`), so the access is safe.
- Matches the SDK base (`openai_compatible/text_embedding.py`) and all sibling embedding providers.

### DIFF
--- a/models/mistralai/manifest.yaml
+++ b/models/mistralai/manifest.yaml
@@ -33,4 +33,4 @@ resource:
     tool:
       enabled: true
 type: plugin
-version: 0.0.10
+version: 0.0.11
--- a/models/mistralai/manifest.yaml
+++ b/models/mistralai/manifest.yaml
@@ -33,4 +33,4 @@ resource:
     tool:
       enabled: true
 type: plugin
-version: 0.0.9
+version: 0.0.10
--- a/models/mistralai/models/text_embedding/text_embedding.py
+++ b/models/mistralai/models/text_embedding/text_embedding.py
@@ -182,5 +182,5 @@ class MistralAITextEmbeddingModel(TextEmbeddingModel):
             price_unit=unit,
             total_price=total_price,
             currency=input_price_info["currency"],
-            latency=time.perf_counter() - self.started_at,
+            latency=time.perf_counter() - getattr(self, "started_at", time.perf_counter()),
         )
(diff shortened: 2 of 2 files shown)


ly-wang19:
Thanks for the review. `self.started_at` is always initialized, so there is no `AttributeError` here: the base `AIModel.__init__` sets `self.started_at = 0` (dify_plugin `interfaces/model/ai_model.py`), making the access safe even when `validate_credentials` calls `_invoke` directly.

This change also matches the SDK's own contract and every sibling embedding provider — e.g. `dify_plugin/interfaces/model/openai_compatible/text_embedding.py` computes `latency=time.perf_counter() - self.started_at`, as do `openai`, `cohere`, `azure_openai`, `voyage`, etc. In the normal `invoke()` path `timing_context()` sets `started_at`; in the `validate_credentials` path the returned `EmbeddingUsage` is discarded, so the latency value there is irrelevant (and this behavior is identical across all providers). Keeping it consistent with the siblings is intentional rather than adding a `getattr` fallback that would diverge from them.

crazywoola (dismissed):

crazywoola (dismissed):

crazywoola (approved):

PR #3331, opened 2026-06-20
fix(bing): build related searches from related_searches, not news

ly-wang19:
## Summary

In the `json` result path, the **related searches** block iterates over `news` instead of `related_searches`:

```python
if related_searches:
    result["related searches"] = [
        {"displayText": item.get("displayText", ""), "url": item.get("webSearchUrl", "")}
        for item in news   # should be related_searches
    ]
```

`news` items carry `name`/`url`, not `displayText`/`webSearchUrl`, so whenever both news and related searches are present the "related searches" output is filled with **empty strings**; and when `news` is empty but related searches exist, the entire "related searches" section silently **disappears**.

Two things confirm the intended behavior:
- The sibling `link` result path (same file) correctly iterates `related_searches` and reads `displayText`.
- `related_searches` is built from `response["relatedSearches"]["value"]`, and the mapped keys (`displayText`, `webSearchUrl`) are relatedSearches fields — so the comprehension was written for `related_searches`; `news` is a copy-paste slip from the `news` block directly above.

Fix: iterate `related_searches`. Version bumped `0.0.10` → `0.0.11`.

## Change Type

- [ ] LLM plugin

## Screenshots / Videos

N/A — JSON output correctness fix.

| Before | After |
| ------ | ----- |
| `"related searches": [{"displayText": "", "url": ""}]` (or missing when news is empty) | `"related searches"` populated from the real `relatedSearches` results |

### DIFF
--- a/tools/bing/manifest.yaml
+++ b/tools/bing/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.10
+version: 0.0.11
 type: plugin
 author: "langgenius"
 name: "bing"
--- a/tools/bing/manifest.yaml
+++ b/tools/bing/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.11
+version: 0.0.12
 type: plugin
 author: "langgenius"
 name: "bing"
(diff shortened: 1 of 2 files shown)


crazywoola (dismissed):

crazywoola (approved):

crazywoola (approved):

PR #3353, opened 2026-06-27
fix(azure_openai): drop unsupported usage detail kwargs

### LINKED ISSUE #3322: [azure_openai] prompt_tokens_details kwargs crash _calc_response_usage — persists in v0.0.59 (re: #2779)
#### Environment
- Dify: **1.14.2**
- Plugin daemon: 0.6.1
- `dify_plugin` SDK in plugin venv: **0.9.0**
- Plugin: **`langgenius/azure_openai` v0.0.59 and v0.0.56**
- Model: any Azure OpenAI LLM used in streaming mode (e.g. gpt-4o)

#### Symptom
Long-running chat / agent / workflow runs that stream an Azure OpenAI LLM **generate content successfully**, then the run **crashes when computing token usage at the very end**. The SSE stream closes with an error **before the message is persisted**, so the conversation appears lost / the page "reboots mid-generation".

#### Error signature
```
PluginInvokeError: [models] Error: LargeLanguageModel._calc_response_usage()
  got an unexpected keyword argument 'prompt_tokens_details'
```

#### Root cause
`models/llm/llm.py` (streaming usage path, ~line 891) calls:
```python
usage = self._calc_response_usage(
    model, credentials, prompt_tokens, completion_tokens,
    prompt_tokens_details=prompt_tokens_details,
    completion_tokens_details=completion_tokens_details,
)
```
But the bundled SDK's `LargeLanguageModel._calc_response_usage()` signature (`dify_plugin/interfaces/model/large_language_model.py:212`) is:
```python
def _calc_response_usage(
    self, model: str, credentials: dict,
    prompt_tokens: int, completion_tokens: int,
) -> LLMUsage:
```
It does **not** accept `prompt_tokens_details` / `completion_tokens_details` → `TypeError` at the end of the stream → workflow node fails → message not persisted.

This is a **version incompatibility**: the plugin was written against a newer SDK that supports these kwargs, but Dify 1.14.2 ships `dify_plugin` 0.9.0, which does not.

This is the same class of bug as #2779, which is **closed but still reproducible** in v0.0.59.

#### Note
- `azure_ai_studio` v0.0.19 is **not** affected (its `_calc_response_usage` call does not pass these kwargs).
- The plugin's other 5 `_calc_response_usage` call sites in the same file already use the safe 4-argument form; only the streaming one passes the kwargs.

#### Minimal fix
Drop the two unsupported kwargs (token totals are still recorded; only the cached/reasoning breakdown is lost):
```python
usage = self._calc_response_usage(
    model, credentials, prompt_tokens, completion_tokens,
)
```

#### Suggested long-term fix
Either (a) gate the kwargs behind an SDK capability/version check, or (b) bump the plugin's minimum `dify_plugin` requirement and document the compatible Dify version, so users on SDK 0.9.0 are not offered an incompatible plugin.

Taranum01:
## Summary
- Stop passing `prompt_tokens_details` and `completion_tokens_details` to `_calc_response_usage` in the Responses API handler
- Aligns with the other five call sites in the same file and with `dify_plugin` 0.9.0 shipped in Dify 1.14.x
- Add regression test ensuring usage calculation uses the 4-argument SDK signature

Fixes #3322

## Context
The Responses API path crashed after a successful generation with:
`LargeLanguageModel._calc_response_usage() got an unexpected keyword argument 'prompt_tokens_details'`

Total token counts are still recorded; cached/reasoning breakdown is omitted until the SDK supports those fields.

## Test plan
- [x] `uv run pytest models/azure_openai/tests/test_llm.py -v` (28 tests pass)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/models/azure_openai/models/llm/llm.py
+++ b/models/azure_openai/models/llm/llm.py
@@ -841,44 +841,12 @@ class AzureOpenAILargeLanguageModel(_CommonAzureOpenAI, LargeLanguageModel):
         # Calculate token usage
         prompt_tokens = 0
         completion_tokens = 0
-        prompt_tokens_details: Optional[dict] = None
-        completion_tokens_details: Optional[dict] = None
 
         if hasattr(response, 'usage') and response.usage:
             usage_obj = response.usage
             # Support Responses API usage fields
             prompt_tokens = getattr(usage_obj, 'input_tokens', None) or getattr(usage_obj, 'prompt_tokens', 0)
             completion_tokens = getattr(usage_obj, 'output_tokens', None) or getattr(usage_obj, 'completion_tokens', 0)
-            # Support implementations that expose detailed fields (convert SDK types to dict)
-            # Prefer the unified prompt_tokens_details/completion_tokens_details fields
-            if hasattr(usage_obj, 'prompt_tokens_details') and usage_obj.prompt_tokens_details:
-                _ptd = usage_obj.prompt_tokens_details
-                if hasattr(_ptd, 'to_dict'):
-                    prompt_tokens_details = _ptd.to_dict()
-                elif isinstance(_ptd, dict):
-                    prompt_tokens_details = _ptd
-                else:
-                    prompt_tokens_details = {
-                        'cached_tokens': getattr(_ptd, 'cached_tokens', None)
-                    }
-            elif hasattr(usage_obj, 'input_tokens_details') and usage_obj.input_tokens_details:
-                it = usage_obj.input_tokens_details
-                if hasattr(it, 'to_dict'):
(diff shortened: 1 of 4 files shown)


crazywoola (changes requested):
https://github.com/langgenius/dify-official-plugins/pull/3354

I suggest including this one in the same PR and do not forget to bump the version in the mainifest.

Taranum01:
Per @crazywoola's review on #3354, the #3322 usage fix is now included there along with a manifest bump to 0.0.60. Closing this PR in favor of #3354.

Taranum01:
Superseded by #3354 (combined azure_openai fixes + manifest 0.0.60).

PR #3363, opened 2026-06-30
fix(linear): accept status name in Update Linear Issue tool

GareArc:
## Problem

The **Update Linear Issue** tool's `New Status` field is documented to take a status name (e.g. "Todo", "In Progress", "Done"), but the code passed the value straight through as `stateId`. So it only worked when users pasted a raw workflow-state UUID — names failed silently.

## Fix

Resolve the provided status to a workflow state ID before the update:

1. Fetch the issue's team (`issue(id){ team { id } }`).
2. List that team's workflow states (`workflowStates(filter: {team:{id:{eq}}}, first: 250)`) — same proven query as `linear_list_issue_statuses`.
3. Match by name (case-insensitive). Fall back to matching a raw state ID for backward compatibility. Unknown name returns an error listing available statuses.

Workflow states are **per-team** in Linear, so resolution is scoped to the issue's own team to stay unambiguous. `first: 250` is Linear's per-page max — covers any team without pagination.

## Changes

- `tools/linear_update_issue.py` — add `_resolve_state_id()`
- `tools/linear_update_issue.yaml` — update `llm_description` to reflect name input
- `manifest.yaml` — version 0.0.7 → 0.0.8

## Notes

Not yet tested against a live Linear workspace.

### DIFF
--- a/tools/linear/tools/linear_update_issue.py
+++ b/tools/linear/tools/linear_update_issue.py
@@ -1,4 +1,4 @@
-from typing import Dict, Any, Generator, Optional, Tuple
+from typing import Dict, Any, Generator
 
 from dify_plugin import Tool
 from dify_plugin.entities.tool import ToolInvokeMessage
@@ -7,6 +7,7 @@ from client.Exceptions import (
     LinearApiException,
     LinearAuthenticationException,
     LinearResourceNotFoundException,
+    LinearValidationException,
 )  # Import standard exceptions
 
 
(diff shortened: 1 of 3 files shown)


tools/linear/tools/linear_update_issue.py
```diff
@@ -13,6 +13,77 @@
...
+    ) -> Tuple[Optional[str], Optional[str]]:
+        """Resolve a status name (or workflow state ID) to a workflow state ID.
+
+        Linear only accepts the workflow state ID. Users normally provide a
+        readable name like "Todo" / "In Progress", so map it to the matching
+        state within the issue's team. A raw state ID is accepted unchanged for
+        backward compatibility.
+
+        Returns:
+            (state_id, None) on success, or (None, error_message) on failure.
```
wylswz: Using `(value, error)` pair adds an error handling path besides try-catch. This is a common paradigm in go, but rarely used in python as far as I know.

wylswz (approved):
LGTM

Tested locally.

Issue #3465, opened 2026-07-20
[searchapi|serper|tianditu] Missing request timeouts + KeyError on missing credentials (mirrors #3456)


Harsh23Kashyap:
## Summary

Three tool plugins (`tools/searchapi`, `tools/serper`, `tools/tianditu`) carry the same three reliability bugs that PR #3456 (already MERGED on 2026-07-20) fixed in `tools/google`. This umbrella issue tracks the rollout across the three remaining plugins.

## The three bugs (recap, same shape as PR #3456)

1. **No `timeout=` on `requests.get`** — a stalled upstream host blocks the worker thread until the framework's `MAX_REQUEST_TIMEOUT=120`, masking the real network problem.
2. **`KeyError` on missing credential** — `self.runtime.credentials["<key>"]` raises `KeyError` instead of yielding a friendly error message when the key is absent or empty.
3. **`_invoke` fall-through on invalid `hl` / `gl`** — some `tools/google`-parallel tools yield an error but do not `return`, so the HTTP request still runs and the user sees a raw upstream error body.

## Affected files

| Plugin | File | Bug 1 (timeout) | Bug 2 (credential) | Bug 3 (hl/gl) |
|---|---|---|---|---|
| searchapi | tools/google.py | yes | yes (`searchapi_api_key`) | yes |
| searchapi | tools/google_news.py | yes | yes | yes |
| searchapi | tools/google_jobs.py | yes | yes | yes |
| searchapi | tools/youtube_transcripts.py | yes | yes | no |
| serper | tools/serper_search.py | yes | yes (`serperapi_api_key`) | no |
| tianditu | tools/geocoder.py | yes | yes (`tianditu_api_key`) | n/a |
| tianditu | tools/staticmap.py | yes | yes | n/a |
| tianditu | tools/poisearch.py | yes | yes | n/a |

`tianditu` plugins do not take `hl` / `gl` parameters — Bug 3 does not apply.

## Reproduction

**Bug 1:** Block the upstream host at the firewall (or point DNS at `127.0.0.1`); invoke the tool. Observe the worker thread hang for up to 120 seconds before timing out.

**Bug 2:** Configure the plugin without an API key (or with an empty key), then invoke the tool. Observe a stack trace ending in `KeyError: '<key_name>'` instead of a user-facing error message.

[...]

PR #3442, opened 2026-07-15
fix(openweather): invoke fall-through, request timeouts, and credential validator status-before-json

Harsh23Kashyap:
## Summary

Fixes #3441.

The `openweather` tool plugin v0.0.8 has four real reliability bugs that are user-visible in any Dify installation: the tool keeps executing the HTTP request after yielding an error message, both outbound `requests.get` calls have no `timeout=`, and the credential validator crashes with `JSONDecodeError` on any non-JSON failure page (e.g. a Cloudflare 502). This PR ships five small commits that fix all four bugs, add a test suite covering the new behavior, and bumps the plugin version.

## Change Type

- [ ] Documentation / non-plugin change
- [x] Non-LLM plugin (tools, extensions, datasource, etc.)
- [ ] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| Empty `city` → HTTP 400 `{"message":"Nothing to geocode"}` | Empty `city` → friendly `"Please tell me your city"`, no HTTP call |
| Missing `api_key` → HTTP 401 `failed:401` | Missing `api_key` → friendly `"OpenWeather API key is required."`, no HTTP call |
| Hung OpenWeather host → 120 s Dify timeout | Hung host → clear error within ~12 s |
| Non-JSON 502 → `JSONDecodeError` stack trace | Non-JSON 502 → `Openweather API rejected the key (HTTP 502): <snippet>` |

## LLM Plugin Checklist

N/A — this is a Non-LLM tool plugin.

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`)
- [x] `dify_plugin>=0.9.0` is declared in `pyproject.toml` and locked in `uv.lock` — unchanged from v0.0.8; the plugin already declared `>=0.9.0` and this PR does not edit `pyproject.toml` or `uv.lock`

Bump: `0.0.8` → `0.0.9` (patch).

## Commits in this PR

| # | SHA | Subject |
| - | --- | ------- |
| 1 | `1e23207e` | `fix(openweather): return after error yields in tool invocation` |
| 2 | `c06d52e4` | `fix(openweather): add 10s timeout to all requests.get calls` |
| 3 | `bed926d2` | `fix(openweather): check response status before parsing json in credential validator` |
| 4 | `2d1f1bab` | `test(openweather): cover invoke fall-through, timeouts, credential validator status-before-json` |
| 5 | `8a4aaa83` | `chore(openweather): bump manifest version to 0.0.9` |

## Testing

[...]

PR #3466, opened 2026-07-20
fix(searchapi): add request timeouts and credential guards

Harsh23Kashyap:
## Summary

Mirrors the three-bug reliability pattern that PR #3456 (already MERGED on 2026-07-20, `tools/google`) just fixed in `tools/google`, but applied to the four files under `tools/searchapi`.

Refs #3465 (umbrella tracker — also includes `tools/serper` and `tools/tianditu`).

### Changes

- `requests.get(...)` now passes `timeout=10` in all four tool files (`google.py`, `google_news.py`, `google_jobs.py`, `youtube_transcripts.py`). 10 s is well above the upstream SearchAPI p99 and well below the framework-level 120 s ceiling.
- `self.runtime.credentials["searchapi_api_key"]` is replaced with an early guard that yields `"SearchAPI API key is required."` and returns; the value is read via `.get()`. No more `KeyError` on a missing or empty key.
- For `google`, `google_news`, and `google_jobs`, invalid `hl` (`ISO 639-1`) / `gl` (`ISO 3166-1 alpha-2`) codes are caught before the HTTP round trip with a friendly error and `return`. A regex-based guard is sufficient — the upstream API does the authoritative validation.
- Every `SearchAPI.run(...)` call site is wrapped in `try` / `except requests.exceptions.RequestException` so a timeout, connection reset, DNS failure, or 5xx surfaces as a friendly message instead of a stack trace.
- New `tools/searchapi/tests/test_searchapi.py` with 16 in-process test cases covering all four tool modules (no `pytest` or `dify_plugin` runtime needed — uses only `unittest.mock` + minimal sys-modules stubs).

## Change Type

## LLM Plugin Checklist

## Version

## Testing

- `python3 -m pytest tests/test_searchapi.py -v` -> **16 passed**.
- `uv run --frozen ruff check .` -> **All checks passed**.
- `python3 -m py_compile tools/google.py tools/google_news.py tools/google_jobs.py tools/youtube_transcripts.py tests/test_searchapi.py` -> OK.

## Risks

- The `hl` / `gl` regex uses `^[A-Za-z]{2}$`; this matches the SearchAPI contract for those parameters (two-letter ISO codes). The upstream API does the final, authoritative validation against supported values.
- Out of scope: `tools/searchapi/uv.lock` is unchanged (no dependency additions).

## Out of scope (deliberate)

- No `tools/utils.py` module extraction — each plugin keeps its own copy of the inline imports. Keeping the diff strictly to the bug fixes.
- No model-provider plugin work; only `tools/*`.

## References

- PR #3456 — the merged precedent in `tools/google`. Same three-fix shape.
- PR #3442 — the previously-merged `tools/openweather` precedent for the same pattern.

### DIFF
--- a/tools/searchapi/manifest.yaml
+++ b/tools/searchapi/manifest.yaml
@@ -42,4 +42,4 @@ tags:
   - news
   - productivity
 type: plugin
-version: 0.0.7
+version: 0.0.8
--- a/tools/searchapi/tests/test_searchapi.py
+++ b/tools/searchapi/tests/test_searchapi.py
@@ -231,7 +231,7 @@ def test_google_invalid_hl_returns_message_and_no_http_call():
     fake_get = mock.Mock(side_effect=AssertionError("requests.get must not be called"))
     with mock.patch.object(google.requests, "get", fake_get):
         messages = _flatten(
-            tool._invoke({"query": "hello", "result_type": "text", "hl": "zzzz", "gl": "us"})
+            tool._invoke({"query": "hello", "result_type": "text", "hl": "12", "gl": "us"})
         )
     assert messages[0][0] == "text"
     assert "Invalid 'hl' parameter" in messages[0][1]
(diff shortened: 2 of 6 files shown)


Harsh23Kashyap:

PR #3468, opened 2026-07-20
fix(serper): add request timeout and credential guard

Harsh23Kashyap:
## Summary

Mirrors the same two-bug reliability pattern that PR #3456 (already MERGED on 2026-07-20, `tools/google`) and PR #3466 (`tools/searchapi`, opened today) just fixed, applied to `tools/serper`.

Refs #3465 (umbrella tracker — also includes `tools/{searchapi, tianditu}`).

### Changes

- `requests.get(...)` now passes `timeout=10`. 10 s is well above the upstream Serper.dev p99 and well below the framework-level 120 s ceiling.
- `self.runtime.credentials["serperapi_api_key"]` is replaced with an early guard that yields `"Serper.dev API key is required."` and returns; the value is read via `.get()`. No more `KeyError` on a missing or empty key.
- Wraps the HTTP call in `try` / `except requests.exceptions.RequestException` so a timeout, connection reset, DNS failure, or 5xx surfaces as a friendly message instead of a stack trace.
- New `tools/serper/tests/test_serper.py` with 6 in-process test cases (no `pytest` or `dify_plugin` runtime needed — uses only `unittest.mock` + minimal sys-modules stubs).

`serper_search.py` does not take user-supplied `hl` / `gl` (the values are hardcoded to `"us"` / `"en"`), so the third bug from PR #3456 (invoke fall-through on invalid `hl` / `gl`) does not apply.

## Change Type

## LLM Plugin Checklist

## Version

## Testing

- `python3 -m pytest tests/test_serper.py -v` -> **6 passed**.
- `uv run --frozen ruff check .` -> **All checks passed**.
- `python3 -m py_compile tools/serper_search.py tests/test_serper.py` -> OK.

## Risks

## Out of scope (deliberate)

- No model-provider plugin work; only `tools/*`.

## References

- PR #3456 — the merged precedent in `tools/google`.
- PR #3466 — the matching `tools/searchapi` PR from this same rollout.

### DIFF
--- a/tools/serper/manifest.yaml
+++ b/tools/serper/manifest.yaml
@@ -37,4 +37,4 @@ resource:
 tags:
   - search
 type: plugin
-version: 0.0.7
+version: 0.0.8
--- a/tools/serper/tests/test_serper.py
+++ b/tools/serper/tests/test_serper.py
@@ -1,0 +1,212 @@
+"""In-process tests for tools/serper (no pytest required).
+
+These tests cover two reliability fixes shipped in this PR:
+
+- Bug 1: missing `timeout=` on `requests.get` calls.
+- Bug 2: `KeyError` when `serperapi_api_key` is missing from credentials.
+
+`serper_search.py` does not take `hl` / `gl` user input (the values are hardcoded
+to "us" / "en"), so the third bug from PR #3456 (invoke fall-through on invalid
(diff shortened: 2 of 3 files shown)


Harsh23Kashyap:

PR #3469, opened 2026-07-20
fix(tianditu): add request timeouts and credential guards

Harsh23Kashyap:
## Summary

Mirrors the same three-bug reliability pattern that PR #3456 (already MERGED on 2026-07-20, `tools/google`) and PR #3466 (`tools/searchapi`) / PR #3468 (`tools/serper`) just fixed, applied to `tools/tianditu`.

Refs #3465 (umbrella tracker).

### Changes

- Every `requests.get(...)` call across the three tool files (`geocoder.py`, `staticmap.py`, `poisearch.py`) now passes `timeout=10`. 10 s is well above the upstream Tianditu API p99 and well below the framework-level 120 s ceiling.
- `self.runtime.credentials["tianditu_api_key"]` is replaced with an early guard that yields `"Tianditu API key is required."` and returns; the value is read via `.get()`. No more `KeyError` on a missing or empty key.
- Wraps all HTTP calls in `try` / `except requests.exceptions.RequestException` so a timeout, connection reset, DNS failure, or 5xx surfaces as a friendly message instead of a stack trace.
- New `tools/tianditu/tests/test_tianditu.py` with 11 in-process test cases (no `pytest` or `dify_plugin` runtime needed — uses only `unittest.mock` + minimal sys-modules stubs).

`tianditu` does not take user-supplied `hl` / `gl` (the geocoder API takes a `keyWord` and `region` from the request body, not as user params), so the third bug from PR #3456 (invoke fall-through on invalid `hl` / `gl`) does not apply.

## Change Type

## LLM Plugin Checklist

## Version

## Testing

- `python3 -m pytest tests/test_tianditu.py -v` -> **11 passed**.
- `uv run --frozen ruff check .` -> **1 pre-existing F401 (`Union` unused in `tools/geocoder.py:2`) is unchanged from `upstream/main`** and unrelated to this PR; all new code added in this PR is clean.
- `python3 -m py_compile tools/geocoder.py tools/staticmap.py tools/poisearch.py tests/test_tianditu.py` -> OK.

## Risks

## Out of scope (deliberate)

- `tools/tianditu/tools/staticmap.py` declares a class named `PoiSearchTool` even though the file is actually a static-map image tool (not POI search). This is a pre-existing naming mistake on `main` and is outside the scope of this bugfix PR.
- The pre-existing `Union` unused import in `tools/geocoder.py:2` is left alone to keep the diff strictly to the bug fixes.

## References

- PR #3456 — the merged precedent in `tools/google`.
- PR #3466 — the matching `tools/searchapi` PR from this same rollout.
- PR #3468 — the matching `tools/serper` PR from this same rollout.

### DIFF
--- a/tools/tianditu/manifest.yaml
+++ b/tools/tianditu/manifest.yaml
@@ -36,4 +36,4 @@ tags:
 - utilities
 - travel
 type: plugin
-version: 0.0.7
+version: 0.0.8
--- a/tools/tianditu/tests/test_tianditu.py
+++ b/tools/tianditu/tests/test_tianditu.py
@@ -1,0 +1,282 @@
+"""In-process tests for tools/tianditu (no pytest required).
+
+These tests cover two reliability fixes shipped in this PR:
+
+- Bug 1: missing `timeout=` on `requests.get` calls.
+- Bug 2: `KeyError` when `tianditu_api_key` is missing from credentials.
+
+`tianditu` does not take user-supplied `hl` / `gl` (the geocoder API takes a
+`keyWord` and an optional `region` already from upstream, not through us),
+so the third bug from PR #3456 (invoke fall-through on invalid hl/gl)
+does not apply.
+
+They use only Python stdlib (`unittest.mock`) and tiny in-test stubs for
+`requests`, `dify_plugin`, and `pydantic` so they can run in environments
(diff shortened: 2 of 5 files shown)


Harsh23Kashyap:

PR #3509, opened 2026-07-25
[gemini] Performance: client cache, FileCache, token fallback, splitter local counting (Fixes #3333)

### LINKED ISSUE #3333: [gemini] Performance: client per-call, broken FileCache, token counting ignores prompt_token_count, splitter API round-trips
## Bug Description

Multiple performance issues in the `gemini` plugin (tested on v0.8.1 and v0.8.4). Each is independent but compounds in multi-node workflows.

---

### Issue 1: `genai.Client` created on every call (LLM + embedding)

**Files:**
- `models/llm/llm.py` — `_generate()` creates `genai.Client(api_key=..., base_url=...)` on every call
- `models/text_embedding/text_embedding.py` — `_invoke()` (L77) and `embed_multimodal()` (L384) each create `genai.Client(api_key=...)` on every call

**Impact:** ~150–400 ms × N calls wasted on TCP+TLS handshakes to `generativelanguage.googleapis.com`.

**Fix:** Module-level client cache with `threading.Lock` (double-checked locking pattern). Cache key = `(api_key, base_url)`. Leave `validate_credentials` with a fresh client.

---

### Issue 2: LLM token counting ignores `usage_metadata.prompt_token_count`

**File:** `models/llm/llm.py`, function `_calculate_tokens_from_usage_metadata`

The function sums only `prompt_tokens_details` and **ignores** `usage_metadata.prompt_token_count`. When `prompt_tokens_details` is absent (some models/configs), `prompt_tokens` defaults to `0`, which triggers the expensive `get_num_tokens()` → `_get_num_tokens_by_gpt2()` fallback — a **pure-Python GPT-2 BPE tokenizer that runs over the entire prompt** (500 ms–2.5 s on long prompts, pegging all CPU cores).

**Fix:** Use `prompt_tokens_standard or usage_metadata.prompt_token_count or 0` instead of just `prompt_tokens_standard`. The GPT-2 fallback only fires when even `prompt_token_count` is absent.

---

### Issue 3: `FileCache` is broken across restarts + race conditions

**File:** `models/llm/utils.py`, class `FileCache`

Three problems:

1. **Disk I/O on every operation**: `exists()`, `get()`, and `setex()` each read+parse the JSON file. `setex()` additionally rewrites the entire file.
2. **Cache key uses `hash()`**: Python's `hash()` is salted per-process (`PYTHONHASHSEED`). After a restart, keys **never match** — persistence is an illusion. Every cache entry is effectively a miss after restart.
3. **No locking**: Concurrent workers read+write the same JSON file without synchronization → lost updates, corrupted JSON (silently resets to `{}`).

**Fix:** Replace with an in-memory `dict` + `threading.Lock` + batched eviction. The Files API upload is free (no cost), so re-upload on restart is acceptable. Persistence was already broken (point 2 above), so in-memory is more correct, not a regression.

---

### Issue 4: Embedding text splitter uses API calls for token counting

**File:** `models/text_embedding/text_embedding.py`, function `_split_texts_to_fit_model_specs`

The recursive text splitter calls `self._count_tokens(client, model, text)` (an **HTTP API round-trip** to Google, ~150–400 ms each) at every recursion level — twice per level.

The class already has `_get_num_tokens_by_gpt2(text)` (local, instant) used by `get_num_tokens()` at L192 — but the splitter doesn't use it.

**Fix:** Replace `_count_tokens(client, model, text)` with `self._get_num_tokens_by_gpt2(text)` in the splitter. GPT-2 BPE **overestimates** vs Gemini's SentencePiece tokenizer (15–30% for English, more for non-Latin), so chunks are conservatively smaller — never exceeds the context limit.

> **Note:** This is compatible with the recursion fix in #3323 — the token counting change is inside the already-fixed splitter.

---

## Environment

- Plugin version: v0.8.1, v0.8.4
- dify_plugin SDK: 0.9.0
- Related: #3323 (embedding recursion/np.average crash fix)

hpiclaranet:
## Summary

Fixes #3333 — four independent performance bottlenecks in the `gemini` plugin, each compounding latency in multi-node workflows.

### Changes

**1. Client caching (LLM + embeddings)**
- Module-level `_get_genai_client()` helper with `threading.Lock` and LRU eviction (max 100 entries), keyed by `(api_key, base_url)`. Reuses the `httpx` connection pool across calls instead of establishing a fresh TCP+TLS handshake (~150–400 ms per call) on every `_generate()` / `_invoke()` / `_invoke_multimodal()` invocation.
- `validate_credentials()` unchanged (intentionally fresh client per credential check).

**2. Token counting fallback (`_calculate_tokens_from_usage_metadata`)**
- When `prompt_tokens_details` is absent for a model/config, the old code returned `prompt_tokens = 0`, which forced the expensive GPT-2 BPE local tokenizer (`_get_num_tokens_by_gpt2`, 500 ms–2.5 s) on every streaming chunk. Now falls back to `usage_metadata.prompt_token_count` — the APIs aggregate total — before falling through to the expensive fallback.

**3. `FileCache` in memory (was: disk-backed JSON + no locking)**
- The original implementation read/parsed (and for `setex`, serialized/wrote) the entire JSON file on every operation with no threading lock — concurrently workers could corrupt the file or lose updates.
- Replaced with an in-memory `dict` + `threading.Lock` + lazy eviction (expired-entry purge at 500 entries / 600s; oldest-entry force-eviction when the purge is insufficient).
- Cache keys in `file_parts.py` already use stable `hashlib.sha256`, so disk persistence would now be functional — but in a containerized plugin-daemon the re-upload cost (Gemini Files API, no financial cost) is negligible vs. the performance and correctness cost of the old I/O-per-call approach. See **Design Decisions** below.

**4. Embedding splitter uses local token estimation**
- `_split_texts_to_fit_model_specs` called `_count_tokens(client, model, text)` — which in the current 0.9.3 code does **two** blocking HTTP requests per recursion level (`client.models.get()` to check supported actions, then `client.models.count_tokens()`).
- Changed to `_get_num_tokens_by_gpt2(text)` — the same local GPT-2 BPE estimator already used by `get_num_tokens()`. GPT-2 overestimates vs Geminis SentencePiece, so resulting chunks are conservatively smaller and never exceed the context limit.

### Design Decisions & Trade-offs

**FileCache persistence:** The disk-backed JSON cache in `utils.py` has been replaced with a pure in-memory cache. The Gemini Files API uploads that this cache backs (via `GeminiFilePartFactory`) are free (no financial cost) and auto-expire on Googles side after ~48h anyway. In a containerized plugin-daemon deployment, disk persistence adds complexity (file permissions, concurrent-write corruption, `/tmp` limits) for negligible benefit — a restart triggers a small number of re-uploads of recently-used files, not correctness loss. The commit message and class docstring document this trade-off explicitly.

**LRU client eviction:** Both client caches (LLM and embedding) use insertion-order reordering on hit to approximate LRU, with a max size of 100 entries and oldest-first eviction when at capacity. This bounds memory/connection-pool growth in multi-tenant deployments where many distinct API keys are active over time.

### Testing

- python3 -m py_compile — pass (3 modified files + conftest + test files)
- `ruff check` — All checks passed
- `ruff format` — all formatted
- `pytest models/tests/` — **203/203 passed** (unit tests + live tests with real GEMINI_API_KEY against Gemini API: streaming, structured output, inline images, tool call ID round-trips, embedding, multimodal embedding)
- Existing `TestSplitTextsTermination` tests updated to mock the new dependency (`_get_num_tokens_by_gpt2` instead of `_count_tokens`); new test `test_splitter_never_calls_live_count_tokens_api` verifies the splitter never performs network calls.
- Autouse conftest fixture `_reset_genai_client_caches` prevents test-suite contamination from module-level caches across tests that reuse the same placeholder API keys.

### Version

manifest.yaml: 0.9.3 → 0.9.4 (PATCH bump — performance and concurrency fixes, no public API changes)

### DIFF
--- a/models/gemini/conftest.py
+++ b/models/gemini/conftest.py
@@ -21,3 +21,28 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
         in_gemini = Path(str(item.path)).resolve().is_relative_to(ROOT)
         if in_gemini and item.get_closest_marker("live") is not None:
             item.add_marker(skipped)
+
+
+@pytest.fixture(autouse=True)
+def _reset_genai_client_caches():
+    """Reset the module-level ``genai.Client`` caches (added for issue #3333)
+    before and after every test.
+
+    ``models.llm.llm._genai_client_cache`` and
+    ``models.text_embedding.text_embedding._genai_emb_client_cache`` are
+    process-wide module-level dicts keyed by API key. Many tests in this
+    suite reuse the same placeholder credentials (e.g. ``"fake-key"``,
+    ``"test-key"``). Without resetting the caches, a test could silently
+    receive a cached client (and its mock) left behind by an earlier,
+    unrelated test instead of the one it just patched -- a correctness
+    hazard specific to the test suite, not to production use (real
+    deployments use real, distinct API keys).
+    """
+    from models.llm import llm as _llm_module
+    from models.text_embedding import text_embedding as _embedding_module
+
+    _llm_module._genai_client_cache.clear()
+    _embedding_module._genai_emb_client_cache.clear()
+    yield
+    _llm_module._genai_client_cache.clear()
+    _embedding_module._genai_emb_client_cache.clear()
--- a/models/gemini/manifest.yaml
+++ b/models/gemini/manifest.yaml
@@ -34,4 +34,4 @@ resource:
     tool:
       enabled: true
 type: plugin
-version: 0.9.3
+version: 0.9.4
--- a/models/gemini/models/llm/llm.py
+++ b/models/gemini/models/llm/llm.py
@@ -2,6 +2,7 @@ import base64
 import json
 import logging
 import re
+import threading
 import time
 from collections.abc import Generator, Iterator, Sequence
 from contextlib import suppress
@@ -46,6 +47,16 @@ from .utils import FileCache
 
 file_cache = FileCache()
 
+# Cache of genai.Client instances keyed by (api_key, base_url). Reusing the
+# client (and its underlying httpx connection pool) avoids a fresh TCP+TLS
+# handshake to generativelanguage.googleapis.com on every LLM node call.
+# Bounded with a simple oldest-first eviction so long-running multi-tenant
+# deployments (many distinct API keys) cannot grow this dict unboundedly and
+# leak open HTTP connection pools.
+_GENAI_CLIENT_CACHE_MAX_SIZE = 100
+_genai_client_cache: dict[tuple[str, Optional[str]], genai.Client] = {}
+_genai_client_cache_lock = threading.Lock()
+
 IMAGE_GENERATION_MODELS = {"gemini-2.5-flash-image", "gemini-3-pro-image-preview"}
 NO_SAMPLING_OR_PREFILL_MODELS = {"gemini-3.6-flash", "gemini-3.5-flash-lite"}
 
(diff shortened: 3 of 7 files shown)


crazywoola:

Please:

PR #3615, opened 2026-08-11
ci: publish merged plugins to staging in a separate step

lin-snow:
## Problem

Staging has received no plugin since it was seeded: **531** plugins against production's **918**, and the newest entry in the `latest` collection is from **2026-01-20**.

It is not only a stale catalog. The backend's `latest` / `latest-tools` rebuild is upload-signal driven — `runSyncLatest` does `GETDEL(signalKey)` and returns immediately when nothing signalled it — so those scheduler tasks have never run on staging, and cannot be tested there until real uploads arrive.

## Change

A second upload step after production, with its own `-u`, split into two steps per @crazywoola's review:

- `continue-on-error`, and it runs after production, so it can neither fail this workflow nor publish something production rejected;
- on failure it writes the reason to the job summary and posts a commit comment — a red merge workflow would invite a re-run, and a re-run uploads the package again;
- `-f`, so that re-run is idempotent;
- both secrets are read into `env` and checked for emptiness, so this is a **no-op until `MARKETPLACE_STAGING_BASE_URL` / `MARKETPLACE_STAGING_TOKEN` exist** — merge order against the secrets does not matter.

Retry is per job, not per step: a failed staging upload heals on the next push of that plugin, or on a backfill run.

Historical plugins are not backfilled by this — it forwards publishes from now on and never reconciles.

### DIFF
--- a/.github/workflows/upload-merged-plugin.yaml
+++ b/.github/workflows/upload-merged-plugin.yaml
@@ -166,30 +166,15 @@ jobs:
         run: |
           uv run --with requests --with dify_plugin python .scripts/uploader -d "${{ matrix.plugin_path }}" -t "${{ secrets.MARKETPLACE_TOKEN }}" --plugin-daemon-path .scripts/dify -u "${{ secrets.MARKETPLACE_BASE_URL }}" -f
 
-      # Staging exists to exercise the publish pipeline -- several backend
-      # scheduler tasks (sync_latest_plugins and the rest of the
-      # upload-signal-driven set) only run when an upload signals them, so a
-      # staging deployment that never receives one cannot be tested at all.
-      #
-      # It must never decide whether a plugin reaches production, so this step
-      # is allowed to fail and reports itself through a commit comment instead
-      # of a red workflow. It also runs after production, so a package
-      # production rejects never lands on staging either.
-      #
-      # -f keeps a re-run idempotent: staging is not the record of truth, and a
-      # 409 from a version some backfill already put there would hide whether
-      # the pipeline reached it.
-      #
-      # Packaging repeats here (the uploader packages what -d points at). The
-      # requirements.txt export above is not redone, and a full security scan
-      # measures 16s, so the second leg costs seconds -- not worth passing the
-      # package between steps for.
+      # Staging cannot be tested without receiving uploads: the backend's
+      # latest/recommended rebuilds only run when an upload signals them.
+      # It must never block a release though -- a red merge workflow invites a
+      # re-run, which uploads the package again. -f keeps that re-run idempotent.
       - name: Upload Plugin (staging)
         id: upload-staging
         continue-on-error: true
         env:
-          # matrix.plugin_path, unlike the community repository's PLUGIN_PATH,
-          # is not in $GITHUB_ENV -- it has to be passed in explicitly.
+          # The matrix value is not in $GITHUB_ENV; pass it in explicitly.
           PLUGIN_PATH: ${{ matrix.plugin_path }}
           STAGING_BASE_URL: ${{ secrets.MARKETPLACE_STAGING_BASE_URL }}
           STAGING_TOKEN: ${{ secrets.MARKETPLACE_STAGING_TOKEN }}
--- a/.github/workflows/upload-merged-plugin.yaml
+++ b/.github/workflows/upload-merged-plugin.yaml
@@ -62,6 +62,11 @@ jobs:
         run: .github/scripts/detect-changed-plugins.sh
 
   upload-merged-plugin:
+    permissions:
+      # contents: write is only what lets the failure report post a commit
+      # comment. The toolkit and the CLI are cloned with
+      # ORG_SCOPE_GITHUB_TOKEN, not with this token.
+      contents: write
     needs: detect-changes
     if: needs.detect-changes.outputs.has_changes == 'true'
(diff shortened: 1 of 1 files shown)


crazywoola (approved):

.github/workflows/upload-merged-plugin.yaml
```diff
@@ -159,4 +159,6 @@ jobs:
 
       - name: Upload Plugin
```
crazywoola, line 160: I think it should be splitted into 2 tasks here.

crazywoola (changes requested):

.github/workflows/upload-merged-plugin.yaml
lin-snow: Done — reworked, and the `--mirror-url` flags are gone from both PRs. The upload is now two steps, production then staging, each with its own `-u`:

```yaml
      - name: Upload Plugin (production)
        run: |
          uv run … .scripts/uploader -d "…" -t "${{ secrets.MARKETPLACE_TOKEN }}" -u "${{ secrets.MARKETPLACE_BASE_URL }}" -f

      - name: Upload Plugin (staging)
        id: upload-staging
        continue-on-error: true
        env:
          PLUGIN_PATH: ${{ matrix.plugin_path }}
          STAGING_BASE_URL: ${{ secrets.MARKETPLACE_STAGING_BASE_URL }}
          STAGING_TOKEN: ${{ secrets.MARKETPLACE_STAGING_TOKEN }}
        run: |
          if [ -z "$STAGING_BASE_URL" ] || [ -z "$STAGING_TOKEN" ]; then
            echo "MARKETPLACE_STAGING_BASE_URL / _TOKEN are not set; skipping the staging upload"
            exit 0
          fi
          uv run … .scripts/uploader -d "$PLUGIN_PATH" -t "$STAGING_TOKEN" -u "$STAGING_BASE_URL" -f

      - name: Report a failed staging upload
        if: steps.upload-staging.outcome == 'failure'
        …posts a commit comment…
```

Per the rest of your note, the staging leg cannot fail this workflow: `continue-on-error: true`, it runs **after** production so a package production rejects never lands on staging, and on failure it posts a **commit comment** rather than turning the run red — this is a `push` event, so there is no PR to comment on. A red merge workflow would invite a re-run, and a re-run uploads the package again.

**You were right about the shape and I was wrong about the reason I resisted it.** I argued a per-target rescan would cost minutes. Measured on a real 2.9 MB package with a `requirements.txt` (`bowenliang123/md_exporter@2.0.0`): a full scan with OSV + PyPI lookups is **16.4s** (`--no-vuln-scan`: 0.1s). Against checkout, cloning the toolkit, downloading the CLI and `uv sync`, that is noise. Two steps also removed a hazard the fan-out had: nothing here needs a new uploader flag, so this PR no longer depends on toolkit `main` moving first — which matters since neither plugin repository pins a toolkit version.

One deliberate choice worth your call: I split into two **steps in one job**, not two **jobs**. Steps are still individually visible and the failing one is obvious, and they reuse the checkout, the toolkit clone, the CLI download and the `requirements.txt` export. Two jobs would repeat all of that (~1–2 min) or need the package plumbed through an artifact. Say the word if you want real separate jobs and I will wire the artifact hand-off.

Also caught while reworking: the staging step first referenced `$PLUGIN_PATH`, which does not exist in this repository — unlike `dify-plugins`, this workflow carries the path in the matrix, not in `$GITHUB_ENV`, so the uploader would have been handed `-d ""`. Fixed by passing `PLUGIN_PATH: ${{ matrix.plugin_path }}` explicitly.

Verification: `actionlint` findings **4 → 4** on both repos (all pre-existing). Both new step scripts were extracted from the YAML and executed with a stubbed `gh`/uploader — unset secrets exit 0 and skip, configured invokes the uploader against staging with `-f`, and the failure path renders the commit comment. So both PRs are a **no-op until `MARKETPLACE_STAGING_BASE_URL` / `MARKETPLACE_STAGING_TOKEN` exist**, and merge order against the secrets does not matter.

crazywoola (dismissed):

lin-snow:
Two Copilot findings, both suppressed so leaving the verdict here. **Both correct**; one was a real bug.

**1. `continue-on-error` only covered the upload — fixed.** The report step called `gh api …/comments` with no `continue-on-error`, so a 403, a rate limit or a transient 5xx would fail it and redden a matrix job whose production publish had already succeeded — inviting the re-run this step exists to prevent (and in `dify-plugins`, production has no `-f`, so that re-run would 409).

Fixed by making the report non-blocking *and* independent of the token: the message goes to `$GITHUB_STEP_SUMMARY` first, which needs no permissions, and the comment is best-effort after it. Verified with a `gh` stub that exits 1 on a simulated 403 — the script exits 1, `continue-on-error` absorbs it, the summary still carries the full report.

**2. "independently re-runnable" was false — claim removed.** More so than stated: GitHub re-runs jobs, not steps, so retrying this job re-runs the production upload too, and in `dify-plugins` that 409s before ever reaching staging. Real independent retry needs the staging leg as a downstream job with the package handed over as an artifact (`-p` needs no plugin-daemon binary, so it stays small). That may be what @crazywoola meant by "2 tasks" — say so and I will wire it; otherwise the description now matches the behaviour.

`actionlint`: still 4 → 4 findings, all pre-existing.

crazywoola (approved):

PR #3630, opened 2026-08-13
fix: force fresh connection for DashScope TTS calls to prevent stale pooled connection hang (closes #3624)

### LINKED ISSUE #3624: tongyi LLM: DashScope 1.26.5+ shared-session reuse can cause 300s ReadTimeouts on stale connections
### Self Checks

### Dify version

1.16.1

### Plugin version

Tongyi 0.2.10 (confirmed reproduction); Tongyi 0.2.11 leaves the affected LLM call path unchanged

### Cloud or Self Hosted

Self Hosted (Docker)

#### Summary

DashScope Python SDK 1.26.5 changed synchronous requests to use a process-lifetime shared `requests.Session`. Tongyi's LLM text-generation path consumes that shared session without an explicit stale-connection validation or recovery strategy before the SDK timeout. In a long-lived Dify plugin worker, this can leave a generation request blocked until DashScope's default 300-second read timeout expires.

This is a compatibility gap between the newer DashScope SDK connection lifecycle and the Tongyi LLM integration. It is **not** a regression introduced by Tongyi 0.2.11: the failure was captured on Tongyi 0.2.10, while 0.2.11 changed the embedding path but left `models/tongyi/models/llm/llm.py` unchanged.

In the captured Agent canary, the initial generation call timed out and produced no tool events. That request remained blocked for approximately five minutes before returning an error.

#### Version and ownership boundary

| Component/version | Relevance |
| --- | --- |
| DashScope SDK 1.26.5 | Introduced the process-lifetime shared synchronous `requests.Session`. |
| DashScope SDK 1.26.6 | Confirmed affected version in the packet-captured reproduction. |
| DashScope SDK 1.26.7 | Retries one pre-response `ConnectionError` on a fresh connection, but explicitly does not retry `ReadTimeout`. |
| Tongyi 0.2.10 | Version used for the confirmed Dify reproduction. |
| Tongyi 0.2.11 | Improves retry and error classification for embeddings; it does not add LLM stale-session recovery. |

The connection-reuse behavior originated upstream in DashScope. The actionable Dify-side gap is that Tongyi's LLM integration has not adapted to that lifecycle change: it has no explicit stale-session strategy and does not surface this silent failure before the SDK's 300-second timeout expires.

This report does not identify which remote or intermediate component discarded the connection state. The evidence is compatible with an idle NAT, firewall, load balancer, or remote endpoint losing state while the local kernel still reports the socket as `ESTABLISHED`; it does not prove which one occurred.

#### Confirmed environment

- Dify 1.16.1, local Docker deployment
- Tongyi plugin 0.2.10
- `dashscope==1.26.6`
- `requests==2.34.2`
- `urllib3==2.7.0`
- Agent model: a DashScope text-generation model invoked through the Tongyi plugin

A separate local Dify 1.14.2 Docker deployment with Tongyi `0.2.9-post731` was also inspected. It contains the same DashScope 1.26.6 process-wide shared-session implementation and 300-second timeout. This second environment confirms exposure to the same code path, but its local logs do not contain a direct reproduction of this exact `ReadTimeout`.

#### Intermittent reproduction

The failure depends on a pooled connection becoming silently stale, so it is intermittent rather than reproducible on every request.

1. Run a long-lived Dify plugin worker using Tongyi 0.2.10 with DashScope 1.26.6.
2. Send Agent-chat requests through `/v1/chat-messages` with `response_mode: streaming` over time, allowing the SDK to reuse its process-wide session.
3. When a pooled connection remains locally `ESTABLISHED` after its state has been lost elsewhere on the path, send another text-generation request.
4. Observe that the request may receive no response and fail only after the 300-second DashScope read timeout.

Diagnostic note: for new conversations, Dify's default `auto_generate_name: true` can add a concurrent title-generation request. This amplified our canary by consuming a second pooled connection, but it is not a prerequisite or the root cause.

#### Packet-level and control evidence

With identical inputs, model configuration, and tool configuration:

- Successful runs completed in approximately 17–22 seconds.
- Failed runs produced no Agent tool events and ended after approximately 300 seconds with `requests.exceptions.ReadTimeout`.
- A metadata-only packet capture started before a failed run observed two pre-existing TCP flows. Neither flow opened with a new SYN.
- Both flows sent application data and repeatedly retransmitted the same sequence ranges, but received no inbound ACK, response, RST, or FIN during the timeout window.
- Fresh control connections from the same host to the same DashScope endpoints completed normally in under 0.4 seconds.
- A subsequent canary with `auto_generate_name: false` used one fresh TCP connection, showed a complete bidirectional exchange, and completed successfully in approximately 22 seconds.

This evidence strongly supports reuse of stale pooled TCP state. It does not establish whether the state was discarded by a local intermediate device, an upstream network component, or the DashScope endpoint.

#### Failed-path sequence

```mermaid
sequenceDiagram
    participant D as Dify Agent
    participant T as Tongyi LLM plugin
    participant S as DashScope SDK
    participant P as Process-wide requests.Session
    participant A as DashScope endpoint

    D->>T: Text-generation request
    T->>S: Generation.call(...)
    S->>P: Acquire a pooled connection
    P-->>S: Existing locally ESTABLISHED socket
    S->>A: POST over the existing connection
    Note over S,A: No new SYN and no reply on the pooled path
    loop Until the 300-second read timeout
        S->>A: TCP retransmission
        A--xS: No ACK or response
    end
    S-->>T: requests.exceptions.ReadTimeout
    T-->>D: Generic plugin invocation failure
    Note over D,A: Control: a fresh TCP connection completed successfully
```

#### Two separate defects exposed together

1. **Session-lifecycle failure:** the LLM path has no explicit stale-session strategy before the SDK's 300-second timeout. In the captured failure, a pooled connection was reused and the invocation remained blocked until that timeout expired. This is the operational failure.
2. **Error misclassification:** `models/tongyi/models/llm/llm.py` defines its own `_invoke_error_mapping`, shadowing the mapping in `models/_common.py`, and omits `requests.exceptions.ConnectionError` and `requests.exceptions.Timeout`. The timeout therefore surfaces as a generic `InvokeError` rather than `InvokeConnectionError`.

Fixing the error mapping would make the failure correctly classified, but it would not shorten the 300-second wait. Session lifecycle and error classification should therefore be addressed and tested separately.

#### Why the existing fixes are insufficient

- [dashscope/dashscope-sdk-python#175](https://github.com/dashscope/dashscope-sdk-python/pull/175), released in DashScope 1.26.7, retries exactly one pre-response `ConnectionError` on a fresh connection. The change explicitly excludes read timeouts.
- Tongyi #3622 / #3623 added retry and error classification to the text-embedding path. The LLM path still calls `Generation.call(...)` / `MultiModalConversation.call(...)` directly, with no stale-session recovery.
- Tongyi 0.2.11 currently declares `dashscope>=1.25.18,<1.26.0` in `pyproject.toml`, while `uv.lock` pins 1.26.6. The declared range excludes DashScope 1.26.7 and is inconsistent with the lockfile.

#### Expected behavior / requested change

Please add a safe connection-lifecycle strategy for Tongyi LLM text generation:

1. Define and test the synchronous session lifecycle for LLM invocations. If an isolated or explicitly managed `requests.Session` is used, keep it valid for the full streaming iteration and ensure that one concurrent invocation cannot close or replace another invocation's active session.
2. Verify that a timed-out pooled connection cannot be reused and that the next invocation obtains healthy TCP state. If the SDK's existing timeout cleanup is insufficient, evict/reset the affected connection pool or session. Do not automatically replay the timed-out generation POST unless the retry is demonstrably safe.
3. Add `requests.exceptions.ConnectionError` and `requests.exceptions.Timeout` to `llm.py`'s own `_invoke_error_mapping` under `InvokeConnectionError`.
4. Reconcile the DashScope dependency declaration with `uv.lock` and test an appropriate supported version, including evaluation of the 1.26.7 pre-response `ConnectionError` fix. A dependency update alone does not handle `ReadTimeout`.
5. Add regression coverage for streaming and concurrent invocations that verifies: a timed-out generation POST is not blindly replayed; the failed connection cannot affect the next invocation; and the LLM error is classified as a connection error.

### ✔️ Error log

Sanitized error excerpt:

```text
PluginInvokeError: [models] Error
requests.exceptions.ReadTimeout:
HTTPSConnectionPool(host='dashscope.aliyuncs.com', port=443):
Read timed out. (read timeout=300)
```

No API keys, internal addresses, request identifiers, user prompts, or business data are included in this report.

botbikamordehai2-sketch:
## What
In TTS calls via the Tongyi plugin, MultiModalConversation.call() reuses DashScope's default HTTP session. When the underlying pooled connection becomes stale (e.g., idle timeout on a firewall), the request can hang for up to the 300-second read timeout, causing failed text generation before reaching any tool or workflow.

## Fix
Disable connection reuse for each TTS call by creating a fresh requests.Session per call and closing it immediately. This prevents stale pooled connections from being reused. The monkey-patch is applied only for the duration of the call to minimize changes.

Closes #3624

### DIFF
--- a/models/tongyi/models/tts/tts.py
+++ b/models/tongyi/models/tts/tts.py
@@ -8,6 +8,30 @@ from dify_plugin.errors.model import (
     InvokeBadRequestError,
     InvokeError,
 )
+import requests
+
+# Force a fresh connection per call to avoid stale pooled connection issues.
+_original_call = MultiModalConversation.call
+def _call_with_fresh_connection(*args, **kwargs):
+    # Create a new session and use it for this single request.
+    # DashScope's HTTP client uses requests.Session internally; we can provide
+    # a custom kwargs to avoid reuse by closing the session after the call.
+    # However, DashScope doesn't expose session param directly, so we use a
+    # context to temporarily set a new session.
+    # As a minimal fix, we monkey-patch the default session used by dashscope's
+    # HTTP client to be a new one each time.
+    # This is done by importing the internal module (subject to change) but
+    # for robustness, we trigger a new connection via a fresh session.
+    import dashscope.common.http_client as http_client
+    original_session = getattr(http_client, '_SESSION', None)
+    http_client._SESSION = requests.Session()
+    try:
+        return _original_call(*args, **kwargs)
+    finally:
+        # Close the session to prevent pooling.
+        http_client._SESSION.close()
+        http_client._SESSION = original_session
+MultiModalConversation.call = _call_with_fresh_connection
 from dify_plugin.interfaces.model.tts_model import TTSModel
 from models._common import _CommonTongyi, get_http_base_address
 


wuyak:
Thanks for taking a look at #3624. I audited this patch against the dependency actually locked by the Tongyi plugin, and the current implementation cannot fix that issue; it will fail before reaching the original TTS call.

### 1. The imported DashScope module does not exist in the locked SDK

Tongyi's [`uv.lock` at the PR base commit](https://github.com/langgenius/dify-official-plugins/blob/4b06a55563df0de5f283caed0061d0b18dab2973/models/tongyi/uv.lock#L403-L420) pins `dashscope==1.26.6`. I checked both the [official PyPI wheel](https://files.pythonhosted.org/packages/2f/9d/d46e56fd96253d34a3bf7140b7abf3cc4e34cf9042f7dcbd57d729fd9f5d/dashscope-1.26.6-py3-none-any.whl) (`sha256:55a333c7b7465831b43b700d41a1209d38f2d952dfe24ca5ff44a3efe8271a87`) and a live Dify Tongyi plugin venv using that version. The following probe is deterministic:

```text
dashscope.__version__ = 1.26.6
find_spec("dashscope.common.http_client") = None
import dashscope.common.http_client
ModuleNotFoundError: No module named 'dashscope.common.http_client'
```

The PR wrapper performs that import on its first invocation ([this commit, `tts.py` lines 11–34](https://github.com/langgenius/dify-official-plugins/blob/ece8b1c84bf7a6d1d33711d316cb9b5ffbf86bef/models/tongyi/models/tts/tts.py#L11-L34)), so the first patched TTS call raises `ModuleNotFoundError` before `_original_call(...)` is reached.

DashScope 1.26.6 defines a lazily created process-shared synchronous session as [`dashscope.api_entities.http_request._shared_sync_session`](https://github.com/dashscope/dashscope-sdk-python/blob/v1.26.6/dashscope/api_entities/http_request.py#L30-L46), and `HttpRequest._handle_request()` reads that session when no external session is supplied ([lines 480–529](https://github.com/dashscope/dashscope-sdk-python/blob/v1.26.6/dashscope/api_entities/http_request.py#L480-L529)). In 1.26.6, writing the `_SESSION` name used by this patch would therefore not control the pool that the SDK actually uses.

The SDK also does support an explicit `session=` kwarg: [`api_request_factory.py` extracts it](https://github.com/dashscope/dashscope-sdk-python/blob/v1.26.6/dashscope/api_entities/api_request_factory.py#L21-L40), and the official generation sample demonstrates passing a custom `requests.Session` ([sample](https://github.com/dashscope/dashscope-sdk-python/blob/v1.26.6/samples/test_generation.py#L337-L371)).

### 2. This patch targets TTS, while #3624 is the LLM text-generation path

The only changed file is `models/tongyi/models/tts/tts.py`. The packet-captured failure in #3624 occurred in the initial LLM generation call. At the PR base commit, the vision branch in `models/tongyi/models/llm/llm.py` uses `MultiModalConversation.call(...)`, while the non-vision text-generation path uses `Generation.call(...)` ([LLM call paths](https://github.com/langgenius/dify-official-plugins/blob/4b06a55563df0de5f283caed0061d0b18dab2973/models/tongyi/models/llm/llm.py#L335-L385)). This PR does not modify the non-vision LLM path, its session lifecycle, or its error mapping.

There is also a scope problem with the monkey patch: `MultiModalConversation.call = _call_with_fresh_connection` is applied at module import and remains process-wide, not only "for the duration of the call" as stated in the PR description. If the TTS module is loaded, it can also intercept other users of `MultiModalConversation.call`; concurrent invocations would additionally race on the same global replacement/state.

### Requested correction

Please do not merge this patch or use it to close #3624 in its current form. At minimum:

1. Remove or replace the nonexistent `dashscope.common.http_client` path.
2. Target the actual LLM call paths and their own `_invoke_error_mapping` for #3624.
3. Prefer the SDK's supported `session=` interface or another tested lifecycle design over a process-wide monkey patch.
4. Add tests for the locked DashScope version, streaming lifetime, concurrent invocations, and the rule that an arbitrary generation `ReadTimeout` must not be blindly replayed.

If this PR is intended to address a separate TTS problem, it should be tracked separately and should not contain `Closes #3624` until it covers the LLM failure reported there.

crazywoola:
Please bump the version in the manifest

wuyak:
Follow-up: I submitted #3659 as a focused fix for #3624 in the actual Tongyi
LLM text and vision paths. It uses DashScope's supported per-invocation
`session=` interface, adds the missing requests error mapping, deliberately
does not replay `ReadTimeout` generation requests, and includes focused
regression tests plus the required version and dependency updates.

Linking it here so maintainers can compare the two implementations and scopes.

PR #3650, opened 2026-08-14
fix(openai_api_compatible): include usage in stream responses

Taranum01:
Fixes #40752

## Problem

OpenAI-compatible servers (vLLM, SGLang, llama.cpp, ...) only emit the `usage` block in the final stream chunk when the client sends `stream_options: {include_usage: true}`. Without it, the base SDK falls back to `_num_tokens_from_string(prompt_messages[0].content)`, which counts only the **first** prompt message.

Reporter example: same request, vLLM returns 56593 prompt_tokens; Dify logs 298.

## Fix

Inject `stream_options: {include_usage: True}` into `model_parameters` before the base SDK call, when `stream=True` and the caller hasn't already supplied their own `stream_options`.

Adds a credential option `stream_include_usage` (default: enabled) so gateways that reject the field can opt out.

## Regression tests

`tests/test_stream_include_usage.py` covers:
- streaming injects the option
- non-streaming skips it
- user-supplied `stream_options` takes precedence
- credential opt-out (`stream_include_usage: disabled`) suppresses the field

All 38 tests pass (4 new + 34 existing).

## Version

Bumps `openai_api_compatible` 0.0.59 → 0.0.60.

### DIFF
--- a/models/openai_api_compatible/manifest.yaml
+++ b/models/openai_api_compatible/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.59
+version: 0.0.60
 type: plugin
 author: "langgenius"
 name: "openai_api_compatible"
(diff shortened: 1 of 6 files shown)


crazywoola:

Please:

PR #3687, opened 2026-08-19
feat(azure_openai): attach Dify app_id as request metadata (opt-in)

ryuta-kobayashi-15:
## Summary

Supersedes #3233. That PR has been open since May and its branch pinned the SDK to a fork that has since been deleted, so it can no longer build. This opens from a fresh branch on current `main` with the pin fixed and review findings addressed. #3233 can be closed once this is reviewed — I don't have permission to close it myself.

Adds an opt-in credential that attaches the current Dify `app_id` to Azure OpenAI requests as `metadata`, so a stored request can be traced back to the Dify app that made it.

Requires `get_current_session()` from the Dify Plugin SDK, released in `dify_plugin` v0.10.0. The API-side plumbing landed in langgenius/dify#35859 and the SDK support in langgenius/dify-plugin-sdks#313.

Rebased onto current `main`. The previous revision pinned the SDK to a personal fork branch that no longer resolves; this revision depends on the released `dify_plugin>=0.10.0`.

## Release Notes

Added an optional **Enable request metadata** provider credential. When enabled, Azure OpenAI calls carry `dify_app_id` and `dify_source` as request `metadata`, so usage can be attributed per Dify app. Because the API only accepts metadata when storage is enabled, turning this on also sets `store=true`, which persists requests and responses on your Azure OpenAI resource. Disabled by default; behaviour is unchanged unless you turn it on.

## Change Type
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
|        |       |

## LLM Plugin Checklist

<details>

</details>

Metadata is attached to the request only; generation behaviour, message flow and token accounting are untouched. Session lookup failures are swallowed so telemetry can never break generation. The storage side effect of `store=true` is documented in the credential help text.

Two guards worth a reviewer's eye, both of which prevent a request the API would reject outright rather than letting it fail generation:

- `openai_api_version` is optional and the fallback default is `2024-02-15-preview`, while the dropdown still offers versions back to `2023-12-01-preview`. Those predate Stored Completions, so metadata is skipped on them instead of failing every call. Versionless `/openai/v1` endpoints are exempt, matching how `common.py` omits `api_version` for them. `validate_credentials` cannot catch this because it builds its own probe request, so without the guard the credential would save cleanly and then break the first real generation.
- An explicit `store=false` from the caller is respected, and metadata is skipped in that case for the same reason.

## Version
- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — `0.0.65` → `0.0.66`
- [x] `dify_plugin>=0.10.0` is declared in `pyproject.toml` and locked in `uv.lock`

`>=0.10.0` is required because `get_current_session()` first ships in that release.

## Testing
- [x] Local deployment — Dify version: 1.16.1
- [ ] SaaS (cloud.dify.ai)

`uv run pytest tests/` — 64 passed against the released SDK v0.10.2.


crazywoola:

Please:

crazywoola (approved):

PR #3688, opened 2026-08-19
feat(openai): attach Dify app_id as request metadata (opt-in)

ryuta-kobayashi-15:
## Summary

Supersedes #3203. That PR has been open since May, its branch pinned the SDK to a fork that has since been deleted, and the 1.0 rewrite (#3419) removed the file it patched — so it can neither build nor merge. This opens from a fresh branch on current `main`, ported to the 1.0 layout, with review findings addressed. #3203 can be closed once this is reviewed — I don't have permission to close it myself.

Adds an opt-in credential that attaches the current Dify `app_id` to OpenAI requests as `metadata`, so a request can be traced back to the Dify app that caused it.

Scope note, to set expectations correctly: `metadata` makes requests filterable on Stored Completions and in Logs. It is not a billing dimension — OpenAI's spend surfaces group by project, key and model, so an organization that wants dollars split per app is still better served by per-project API keys. What metadata gives you is the per-request join back to a Dify app, which project-level keys cannot express when one key serves many apps.

Requires `get_current_session()` from the Dify Plugin SDK, released in `dify_plugin` v0.10.0. The API-side plumbing landed in langgenius/dify#35859 and the SDK support in langgenius/dify-plugin-sdks#313.

Rebased onto current `main` and **ported to the 1.0 rewrite** (#3419), which replaced the single `llm.py` the original patch targeted. The insertion points are now:

- `chat.py` — in `generate_chat`, right after `_add_identity`, where credentials are already in scope.
- `responses.py` — `parameters()` takes an optional `credentials` argument, applied before the `store`/`include` handling because the API only accepts `metadata` when `store` is true. Existing four-argument call sites keep working unchanged.
- `stream.py` — credentials threaded through the streaming call site.

Reasoning behaviour is deliberately preserved: when `store=true` was set only to carry telemetry, `reasoning.encrypted_content` is still requested, so opting in does not silently change multi-turn reasoning on gpt-5 / o-series models.

The previous revision pinned the SDK to a personal fork branch that no longer resolves; this revision depends on the released `dify_plugin>=0.10.0`.

## Release Notes

Added an optional **Enable request metadata** credential, available at both provider and model scope. When enabled, `chat.completions.create` and `responses.create` calls carry `dify_app_id` and `dify_source` as request `metadata`, so a stored request can be traced back to the Dify app that made it. Because the API only accepts metadata when storage is enabled, turning this on also sets `store=true`, which persists requests and responses to Stored Completions on your OpenAI account. Disabled by default; behaviour is unchanged unless you turn it on.

## Change Type
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
|        |       |

## LLM Plugin Checklist

<details>

</details>

Both request paths (Chat Completions and Responses) are covered. Encrypted reasoning passthrough is retained when `store=true` comes from telemetry alone, so reasoning continuity is unchanged. An explicit `store=false` from the caller is respected and metadata is skipped rather than emitting a request the API would reject. Session lookup failures are swallowed so telemetry can never break generation.

## Version
- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — `1.0.3` → `1.0.4`
- [x] `dify_plugin>=0.10.0` is declared in `pyproject.toml` and locked in `uv.lock`

`>=0.10.0` is required because `get_current_session()` first ships in that release. `pyproject.toml` and the `test_catalog.py` version assertion were bumped alongside `manifest.yaml`.

## Testing
- [x] Local deployment — Dify version: 1.16.1
- [ ] SaaS (cloud.dify.ai)

`uv run pytest tests/` — 184 passed, 65 live tests skipped (no API key), against the released SDK v0.10.2.


crazywoola:

Please:

crazywoola (approved):

Issue #3658, opened 2026-08-15
anthropic: thinking blocks are not preserved across plugin invocation boundaries on tool-result turns


hpiclaranet:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [x] I have searched for existing issues [Dify issues](https://github.com/langgenius/dify/issues) & [Dify Official Plugins](https://github.com/langgenius/dify-official-plugins/issues), including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.14.2 (self-hosted, docker-compose)

### Plugin version

langgenius/anthropic 0.3.26 (current main @ eca4e125)

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

**Summary**

The plugin stores Claude's `thinking` / `redacted_thinking` blocks from a response on the model **instance** and re-injects them when building the next request of a tool-use loop:

```python
# __init__
self.previous_thinking_blocks = []
self.previous_redacted_thinking_blocks = []

# _process_assistant_message (next-turn request building)
if has_tool_messages:
    content.extend(self.previous_thinking_blocks)
    content.extend(self.previous_redacted_thinking_blocks)
```

Because the SDK builds a **fresh model instance for every invocation** (`dify_plugin` **0.10.0**: [`core/model_factory.py#get_instance`](https://github.com/langgenius/dify-plugin-sdks/blob/0.10.0/python/src/dify_plugin/core/model_factory.py#L38) returns `self.models[model_type](self.provider.models)`; called per RPC from `plugin_registration.py:get_model_instance()`, ~10 handler sites in `plugin_executor.py`), this state is always empty by the time the follow-up request is constructed.

The assistant message of the tool turn is therefore sent **without** the thinking blocks that Anthropic requires to be passed back unmodified within a tool-use turn ([thinking docs — preserving blocks](https://docs.claude.com/en/docs/build-with-claude/thinking)). Notably, with `thinking.type=adaptive` + `display: "omitted"` (the default on Opus 5 / Sonnet 5) the response still contains a `thinking` block — readable text empty, but the cryptographic `signature` present and required on the round-trip.

**Impact / when it fires**

[...]

PR #3691, opened 2026-08-19
fix(anthropic): persist thinking blocks on the message, not the model instance

AmirF194:
## Summary

`dify_plugin`'s `ModelFactory.get_instance()` builds a fresh `AnthropicLargeLanguageModel` instance for every RPC (its own docstring: "generate stateless model instances"), and `plugin_registration.get_model_instance()` is called from every handler in `plugin_executor.py`. This plugin stores a response's thinking/`redacted_thinking` blocks on `self.previous_thinking_blocks` / `self.previous_redacted_thinking_blocks`, then reads them back in `_process_assistant_message` while building the next request. The follow-up tool-result turn runs on a different instance, so those lists are always empty by the time they're read: Anthropic requires the blocks to be echoed back verbatim on that turn, and never gets them.

Confirmed by exercising `ModelFactory.get_instance()` (`self.models[model_type](self.provider.models)`, a fresh call each time) directly, and by driving `_handle_chat_generate_response` / `_process_assistant_message` from two separate `AnthropicLargeLanguageModel()` instances with a synthetic thinking + tool_use response.

## What this changes

Moves the state onto `AssistantPromptMessage.opaque_body`, the field Dify persists with the message itself, so it survives whichever instance handles the next turn. This is the same pattern `models/moonshot/models/llm/llm.py` already uses for `reasoning_content`. Both the non-streaming and streaming response handlers now attach the blocks to the message they return instead of to `self`; the `self.previous_*` resets, which existed only to guard against same-instance reuse across unrelated conversations, are removed as dead code now that nothing reads them.

## Not included, and why

Scoped to the in-memory, same-session tool-loop case this issue reports. Persistence across a chat reload or workflow pause-resume, and signature invalidation on a model switch, are open questions the issue itself raises; this PR does not attempt to answer them, since I can't verify Dify core's persistence layer end-to-end from this side of the plugin boundary.

## Release Notes

Fix: extended-thinking tool calls on Anthropic models no longer lose their thinking blocks on the follow-up turn, which previously caused the request to be rejected or degraded whenever Dify routed it to a new plugin instance (nearly always, in production).

## Change Type

- [x] LLM plugin

## Screenshots / Videos

N/A, backend fix with no UI surface.

<details>

</details>

## Version

- [x] Bumped `version` in `manifest.yaml`: `0.3.27` -> `0.3.28`.
- [x] `dify_plugin>=0.9.0` remains declared in `pyproject.toml` and locked in `uv.lock` (unchanged by this PR).

## Testing

- [x] New regression test drives `_handle_chat_generate_response` on one instance and `_process_assistant_message` on a second, separate instance (mirroring `ModelFactory.get_instance()`), and confirms the thinking block survives the boundary. Fails on `main`, passes on this branch.
- [x] `uv run --frozen pytest`: 28 passed, 14 skipped (skipped suite needs a live `ANTHROPIC_API_KEY`), in a clean `python:3.12-slim` container matching this plugin's `requires-python`.
- [x] `uv run --frozen pyrefly check`: 0 errors, same as `main`.
- [ ] Not tested against a running Dify server (local or SaaS); the mechanism is verified at the plugin/message level, not end-to-end through Dify's own conversation persistence.

- [ ] Local deployment
- [ ] SaaS (cloud.dify.ai)

Refs #3658

### DIFF
--- a/models/anthropic/manifest.yaml
+++ b/models/anthropic/manifest.yaml
@@ -25,4 +25,4 @@ resource:
     model:
       enabled: false
 type: plugin
-version: 0.3.27
+version: 0.3.28
--- a/models/anthropic/models/llm/llm.py
+++ b/models/anthropic/models/llm/llm.py
@@ -215,8 +215,6 @@ class AnthropicLargeLanguageModel(LargeLanguageModel):
 
     def __init__(self, model_schemas=None):
         super().__init__(model_schemas or [])
-        self.previous_thinking_blocks = []
-        self.previous_redacted_thinking_blocks = []
         # Flag to indicate whether tool definitions should include cache_control
         self._tool_cache_enabled = False
         self._system_cache_enabled = False
@@ -659,10 +657,6 @@ class AnthropicLargeLanguageModel(LargeLanguageModel):
             else:
                 extra_headers["anthropic-beta"] = "pdfs-2024-09-25"
 
-        if not any(isinstance(msg, ToolPromptMessage) for msg in prompt_messages):
-            self.previous_thinking_blocks = []
-            self.previous_redacted_thinking_blocks = []
-
         if tools:
             extra_model_kwargs["tools"] = [
                 self._transform_tool_prompt(tool) for tool in tools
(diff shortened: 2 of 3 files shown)


crazywoola:

Please:

PR #3704, opened 2026-08-21
[gemini] Performance: client connection reuse, in-memory FileCache, local token counting (Fixes #3333)

### LINKED ISSUE #3333: [gemini] Performance: client per-call, broken FileCache, token counting ignores prompt_token_count, splitter API round-trips

hpiclaranet:
## Summary

Follow-up to #3333 (auto-closed) and supersession of #3509 (closed during the plugin governance initiative, per maintainer request to resubmit against current `main` with the new PR template).

`gemini` 0.9.5 still pays several per-call overheads on every LLM / embedding node invocation in multi-node workflows (RFP analysis, agent loops):

1. **`genai.Client` constructed per call** — a fresh `httpx.Client` (empty connection pool) is built inside `_generate()` and `_generate_via_interactions()` (LLM) and inside `_invoke()` and `embed_multimodal()` (embedding). Every node call pays a full TCP+TLS handshake to `generativelanguage.googleapis.com` (~150–400 ms × N nodes per workflow).
2. **Disk-backed, unlocked `FileCache`** — the Files-API upload cache re-reads and re-writes the whole JSON file on *every* `exists`/`get`/`setex`, with no locking (concurrent workers can lose updates or corrupt the file), and pays blocking disk I/O on every multimodal content check.
3. **API round-trips inside the embedding text splitter** — `_split_texts_to_fit_model_specs` called `client.models.count_tokens()` (blocking HTTP, 150–400 ms) at every recursion level, even though a local GPT-2 estimator (`_get_num_tokens_by_gpt2`) already exists on the class.
4. **`prompt_token_count` ignored in usage mapping** — when the per-modality breakdown is absent, `prompt_tokens` collapsed to 0, forcing the expensive local GPT-2 BPE re-count of the whole prompt elsewhere.

## Release Notes

- Performance: LLM and embedding calls now reuse a single `genai.Client` (and its HTTP connection pool) per credential, eliminating per-call TCP+TLS handshakes to the Gemini API.
- Performance: the Files-API upload cache is now an in-memory, thread-safe cache with lazy expiry-based eviction (replaces a disk JSON file that was rewritten on every access; entries no longer survive a process restart, which only costs a free re-upload).
- Performance: the embedding text splitter now uses the local token estimator instead of a blocking `count_tokens` API call per recursion level.
- Performance: token-usage mapping now falls back to the API-reported `prompt_token_count` when the per-modality breakdown is absent.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

Not applicable — no UI-facing behavior change (performance only; identical model outputs).

## LLM Plugin Checklist

<details>

- [x] Multimodal input (images, PDFs, audio, video, etc.) — FileCache backs the Files-API upload cache used for multimodal parts
- [x] Token consumption metrics — `prompt_token_count` fallback in usage mapping
- [x] Other LLM functionality (reasoning, grounding, prompt caching, etc.) — client reuse in `_generate` and `_generate_via_interactions`

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — `0.9.5` → `0.9.6`
- [x] SDK dependency declared in `pyproject.toml` (`dify-plugin>=0.9.0`) and locked in `uv.lock` — unchanged by this PR

## Testing

- [x] Local deployment — Dify version: 1.14.2 (self-hosted production; the exact same changes have been running in this environment for several weeks on gemini 0.9.5 with no regressions observed)
- Unit tests (new, `models/tests/test_perf_caches.py`, 19 offline tests — no API key required, collected by the plugin's default `testpaths`, run in CI):
  - LLM/emb client-cache identity (same creds → same instance; distinct creds → distinct instances; exactly 1 construction per credential; LLM and embedding caches independent)
  - Regression guard: the LLM class body no longer constructs `genai.Client` outside `validate_credentials` (source-level guard, scoped to the current layout)
  - `FileCache`: setex/get/expiry, 8-thread × 200-key concurrent writes with zero lost updates, concurrent reads during writes, eviction purges expired entries only (live entries survive), backward-compatible constructor signature
  - Usage mapping: modality breakdown preferred, `prompt_token_count` fallback, all-missing → zero
  - Splitter: never invokes API token counting (monkeypatched to fail), short-text pass-through, CJK token-dense text terminates
- Existing suite: added an autouse fixture in the plugin's root `conftest.py` clearing the new module-level client caches per test (existing tests patch `genai.Client` and expect their fresh mock); the splitter tests in `test_embedding.py` now patch the splitter's actual token source (`_get_num_tokens_by_gpt2`) so their determinism (including the zero-token-count guard) is preserved.

Behavioral notes / accepted trade-offs:
- The Files-API upload is free, so the cache losing entries on process restart (cost: one free re-upload per distinct file) is an accepted trade-off for removing the unlocked per-access disk I/O.
- GPT-2 typically *overestimates* vs Gemini's SentencePiece tokenizer, so splitter chunks tend conservatively smaller. This is a heuristic, **not a formal upper bound**, for CJK/emoji/code-heavy text (accepted for the ~150–400 ms per-level saving); the in-code comment states this explicitly.
- Known limitation (pre-existing, unchanged by this PR): the check→upload→set sequence around the cache is not atomic, so two concurrent requests for the same payload can upload twice. A per-key single-flight is a candidate follow-up.
- Known limitation (documented assumption): the client caches are unbounded (one client per distinct credential for the process lifetime), matching the pattern already merged upstream in `azure_openai` (`models/azure_openai/models/common.py`). In practice the cardinality is the number of distinct Gemini credentials configured in the tenant(s) served by the plugin process.
- `validate_credentials` intentionally keeps constructing a fresh client (correct for validation semantics).

### DIFF
--- a/models/gemini/conftest.py
+++ b/models/gemini/conftest.py
@@ -4,8 +4,28 @@ from pathlib import Path
 import pytest
 from dotenv import load_dotenv
 
+from models.llm import llm as gemini_llm
+from models.text_embedding.text_embedding import _genai_emb_client_cache
+
 
 ROOT = Path(__file__).resolve().parent
+
+
+@pytest.fixture(autouse=True)
+def _clear_genai_client_caches():
+    """Isolate the plugin's module-level genai.Client caches between tests.
+
+    The plugin reuses one genai.Client per credential for connection-pool
+    performance (see models/llm/llm.py and models/text_embedding/
+    text_embedding.py). Tests that patch ``genai.Client`` expect their freshly
+    configured mock to be the one in use, so the caches must be empty before
+    and after every test.
+    """
+    gemini_llm._genai_client_cache.clear()
+    _genai_emb_client_cache.clear()
+    yield
+    gemini_llm._genai_client_cache.clear()
+    _genai_emb_client_cache.clear()
 
 
 def pytest_configure() -> None:
--- a/models/gemini/manifest.yaml
+++ b/models/gemini/manifest.yaml
@@ -34,4 +34,4 @@ resource:
     tool:
       enabled: true
 type: plugin
-version: 0.9.5
+version: 0.9.6
--- a/models/gemini/models/llm/llm.py
+++ b/models/gemini/models/llm/llm.py
@@ -2,6 +2,7 @@ import base64
 import json
 import logging
 import re
+import threading
 import time
 from collections.abc import Generator, Iterator, Sequence
 from contextlib import suppress
@@ -43,6 +44,28 @@ from google.genai import errors, types
 from .file_parts import GeminiFileMode, GeminiFilePartFactory
 from .model_schema import with_inline_file_parameter
 from .utils import FileCache
+
+_genai_client_cache: dict[tuple[str, Optional[str]], genai.Client] = {}
+_genai_client_cache_lock = threading.Lock()
+
+
+def _get_genai_client(api_key: str, base_url: Optional[str] = None) -> genai.Client:
+    """Return a cached genai.Client for the given credentials, creating one if needed.
+
+    Reusing the client (and its underlying httpx connection pool) avoids a fresh
+    TCP+TLS handshake to generativelanguage.googleapis.com on every LLM node call.
+    """
+    cache_key = (api_key, base_url)
+    with _genai_client_cache_lock:
+        client = _genai_client_cache.get(cache_key)
+        if client is None:
+            client = genai.Client(
+                api_key=api_key,
+                http_options=types.HttpOptions(base_url=base_url),
+            )
+            _genai_client_cache[cache_key] = client
+        return client
+
 
 file_cache = FileCache()
 
@@ -220,7 +243,7 @@ class GoogleLargeLanguageModel(LargeLanguageModel):
         # The reasoning content and final answer of the Gemini model are priced using the same standard.
         completion_tokens = thoughts_token_count + candidates_token_count
         # The `prompt_tokens` includes the historical conversation QA plus the current input.
-        prompt_tokens = prompt_tokens_standard
+        prompt_tokens = prompt_tokens_standard or usage_metadata.prompt_token_count or 0
 
         return prompt_tokens, completion_tokens
 
@@ -1225,11 +1248,8 @@ class GoogleLargeLanguageModel(LargeLanguageModel):
         # == InitConfig == #
 
         config = types.GenerateContentConfig()
-        genai_client = genai.Client(
-            api_key=credentials["google_api_key"],
-            http_options=types.HttpOptions(
-                base_url=credentials.get("google_base_url", None)
-            ),
+        genai_client = _get_genai_client(
+            credentials["google_api_key"], credentials.get("google_base_url", None)
         )
 
         # == ChatConfig == #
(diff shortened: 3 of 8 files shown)


crazywoola:
Please resolve the conflicts.

hpiclaranet:
Rebased onto current main and reopened as #3759 (GitHub blocks reopening a closed PR after a force-push to its branch): https://github.com/langgenius/dify-official-plugins/pull/3759

Issue #3705, opened 2026-08-21
[cot_agent] ReAct stream parser: adjacent JSON roots, nested function envelopes, and split think tags are misparsed


hpiclaranet:
### Self Checks

- [x] This is only for bug report, if you would like to ask a question, please head to Discussions.
- [x] I have searched for existing issues [Dify issues](https://github.com/langgenius/dify/issues) & [Dify Official Plugins](https://github.com/langgenius/dify-official-plugins/issues), including closed ones.
- [x] I confirm that I am using English to submit this report.
- [x] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [x] Please do not modify this template :) and fill in all the required fields.

### Dify version

Any — this is a plugin-level parser defect (version-independent)

### Plugin version

0.0.45 (current `main`; the affected streaming state machine in `cot_output_parser.py` is unchanged across recent releases, including the 0.0.40 reported in #3699)

### Cloud or Self Hosted

- Cloud
- Self Hosted (Docker)

### Steps to reproduce

Three edge-case defects in the ReAct stream parser (`agent-strategies/cot_agent/output_parser/cot_output_parser.py`, `CotAgentOutputParser.handle_react_stream_output`), found while reviewing PR #3703 (fix for #3699). All live in the same streaming state machine and were intentionally left out of that PR as separate follow-ups.

**1. Adjacent JSON roots drop the first blob**

When the model emits two JSON objects with no delimiter between them, e.g.:

```
{"action": "first", "action_input": {}}{"action": "second", "action_input": {}}
```

only the last blob is processed. When the root of the first blob closes, `got_json` is set and the loop `continue`s; the very next character (the second `{`) enters the "start new JSON" branch, which resets `got_json = False` and overwrites `json_cache` before the first blob was ever parsed. The first (valid) action is silently discarded, so the round ends without invoking `first`.

Expected: both blobs are processed (at minimum, the first action must not be lost).

**2. Two-level OpenAI-style function envelopes are not unwrapped**

```
{"tool_call": {"function": {"name": "webSearch", "arguments": {"query": "x"}}}}
```

[...]

PR #3706, opened 2026-08-21
fix(cot_agent): ReAct parser edge cases from #3705 (adjacent JSON, function envelopes, split think tags)

hpiclaranet:
## Summary

Fixes the three edge cases tracked in #3705 in the `cot_agent` ReAct output parser (`CotAgentOutputParser`):

1. **Adjacent JSON roots** — `{"action": "first", ...}{"action": "second", ...}` (no delimiter between roots) dropped the first blob: a completed blob was only processed at the *bottom* of the per-character loop, by which time the next `{` could already have overwritten `json_cache`. A completed blob is now flushed **before** the next character is accepted.
2. **OpenAI-style function envelopes** — `{"function": {"name": ..., "arguments": ...}}` (bare, wrapped in a single-key container such as `tool_call`, optionally alongside a `"type": "function"` key) is now normalized to a canonical action before extraction. Explicit canonical name keys (`action`, `tool`, `tool_name`, `name`) still take precedence over the envelope, and a `function` value that is not a dict with a string `name` leaves the payload untouched.
3. **Think tags split across stream chunks** — the think-tag stripper (for `include_thoughts` models) only matched complete tags inside a single chunk. A chunk tail that could still be the start of a tag is now held back until the next chunk (and flushed at stream end), so tags split across chunk boundaries are stripped as well.

> [!IMPORTANT]
> **This PR is stacked on #3703** (same file — both touch the ReAct output parser). Please review it as the incremental diff on top of #3703 and merge it after #3703. Once #3703 is merged I will rebase, and the diff against `main` will then show only these changes.

## Release Notes

- ReAct: fix silently dropped tool calls when the model emits two JSON action blocks with no delimiter between them
- ReAct: recognize OpenAI-style function envelopes (`{"function": {"name": ..., "arguments": ...}}`) as tool calls
- ReAct: strip model think tags even when they are split across stream chunks (prevents leaked tag fragments in thoughts/answers)

## Change Type

- [ ] LLM plugin

## Screenshots / Videos

Not applicable — parser behavior only, covered by the unit tests below.

| Before | After |
| ------ | ----- |
|        |       |

## LLM Plugin Checklist

Not applicable — this is an agent strategy plugin, not an LLM provider plugin.

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — 0.0.46 → 0.0.47
- [x] `dify_plugin>=0.3.0,<0.6.0` is declared in `pyproject.toml` and locked in `uv.lock`

## Testing

- [x] Local deployment — clean `uv` venv aligned with `manifest.yaml`, `pyproject.toml`, and `uv.lock`: **76 unit tests pass**, including 13 new regression tests covering each fixed case (adjacent roots in one chunk and across chunks, non-action adjacent blobs, envelope bare/wrapped/typed/no-arguments/null-arguments/empty-name, explicit-key precedence, case-insensitive wrapper keys, think tags split on the opening tag, the closing tag, and a trailing partial tag at stream end); `ruff check` clean (no new findings)
- [ ] SaaS (cloud.dify.ai)

Known limitation (unchanged by this PR, noted in #3705): the think-tag stripper is not JSON-string-aware, so a literal complete think tag inside a JSON string value is stripped, exactly as it already was before this change.

### DIFF
--- a/agent-strategies/cot_agent/manifest.yaml
+++ b/agent-strategies/cot_agent/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.46
+version: 0.0.47
 type: plugin
 author: "langgenius"
 name: "agent"
--- a/agent-strategies/cot_agent/output_parser/cot_output_parser.py
+++ b/agent-strategies/cot_agent/output_parser/cot_output_parser.py
@@ -84,7 +84,7 @@ def parse_action(json_str: str) -> "AgentScratchpadUnit.Action | None":
     # Unwrap recognized single-key wrappers, e.g. {"tool_call": {...}}.
     if len(payload) == 1:
         key, value = next(iter(payload.items()))
-        if key in WRAPPER_KEYS and isinstance(value, (dict, list)):
+        if key.lower() in WRAPPER_KEYS and isinstance(value, (dict, list)):
             if isinstance(value, list):
                 if len(value) != 1:
                     return None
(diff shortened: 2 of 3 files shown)


crazywoola (dismissed):

crazywoola:
Please resolve the conflicts

hjpinheiro:
Resolved: the #3703 changes are already in upstream main as 2ee07344 (co-author credit preserved), so I removed the duplicate commit and rebased the #3705 changes onto current main. The PR now contains only the three-file ReAct parser edge-case diff; CI will rerun on the new head (8aab224d). Since the review approval was tied to the previous head, a re-approval would unblock the merge.

hpiclaranet:
Resolved: the #3703 changes are already in upstream main as 2ee07344 (co-author credit preserved), so I removed the duplicate commit and rebased the #3705 changes onto current main. The PR now contains only the three-file ReAct parser edge-case diff; CI will rerun on the new head (8aab224d). Since the review approval was tied to the previous head, a re-approval would unblock the merge.

crazywoola (approved):

PR #3714, opened 2026-08-21
feat(openai_api_compatible): add video_transfer_format and audio_transfer_format credentials (issue #3696)

### LINKED ISSUE #3696: openai_api_compatible: video attachments are sent as `image_url`, so OpenAI-compatible servers that implement `video_url` reject them
### Summary

With `video_support: support`, attaching a video in Dify fails against OpenAI-compatible servers that implement the `video_url` content part (for example vLLM, directly or behind a LiteLLM proxy). The plugin converts `PromptMessageContentType.VIDEO` into an `image_url` part carrying a `data:video/mp4;base64,...` URI, and the server tries to decode that as an image.

### Steps to reproduce

1. Register a vLLM-served multimodal model (Gemma 3 / Qwen VL class model) through an OpenAI-compatible endpoint.
2. Set `vision_support: support` and `video_support: support` on the model credential.
3. Send a chat message with a short mp4 attached.

### Result

The request fails with HTTP 400 from the serving side:

```
Failed to load image: cannot identify image file <_io.BytesIO object>
```

Verified two ways: through Dify's own invocation path (`PluginModelClient.invoke_llm` with a `VideoPromptMessageContent`), and by posting the exact payload shape the plugin produces to the gateway.

### Expected

The same file is read correctly when the part is sent as `video_url` instead. Measured with a 3 second mp4 whose three 1 second segments are filled with green, magenta and yellow, asking for the colour order:

| content part the plugin sends | result |
| --- | --- |
| `image_url` with `data:video/mp4;base64,...` (current) | HTTP 400 `Failed to load image` |
| `video_url` with the same data URI | the three colours are answered in order |

This held for 12 model names across three vLLM deployments (vLLM 0.26.0 and 0.27.1), and for four models on a hosted OpenAI-compatible provider that also runs vLLM. Ollama and SambaNova backends do not accept video in either shape, so they are unaffected by this report.

### Why the current shape exists

#3090 chose the `image_url`-with-data-URI shape because LiteLLM dispatches it into Vertex Gemini's `inline_data`, and reported that `video_url` and `input_audio` were silently dropped on LiteLLM's Gemini route. That reasoning is provider specific. vLLM defines `video_url` for video input and LiteLLM forwards it as-is for `hosted_vllm`, so on those backends the current shape can never decode.

### Proposal

Make the wire format selectable rather than fixed. For example a `video_transfer_format` credential with `image_url_data_uri` (today's behaviour, default) and `video_url`, and the equivalent for audio with `input_audio`. That keeps Gemini-via-LiteLLM working while letting OpenAI-compatible servers receive the shape they implement.

I am happy to send a PR along those lines if the direction is acceptable.

### Environment

- Plugin: `langgenius/openai_api_compatible` 0.0.59 in use. 0.0.61 and `main` (192fb7e) carry the same implementation
- Dify: 1.16.1, self-hosted with Docker
- Serving: LiteLLM 1.97.0 in front of vLLM 0.26.0 / 0.27.1

Harsh23Kashyap:
## Summary

Adds two new LLM-scope credentials to the `openai_api_compatible` plugin: `video_transfer_format` and `audio_transfer_format`. They let users on OpenAI-compatible servers that implement the OpenAI multimodal spec (vLLM, LiteLLM `hosted_vllm`) send VIDEO content as a `video_url` part and AUDIO content as an `input_audio` part, instead of the current fixed `image_url` data URI that those servers reject with HTTP 400.

The default value for both is `image_url_data_uri` (today's behaviour), so the change is **strictly additive** — no existing user is affected unless they opt in.

4 files, +329/-15, manifest `0.0.63 → 0.0.65`.

## Problem

Issue #3696 documents that VIDEO and AUDIO content are hard-coded to be sent as an `image_url` part with a data URI. This works through LiteLLM to Vertex Gemini (which dispatches to `inline_data`), but OpenAI-compatible servers that implement the OpenAI multimodal spec expect a `video_url` part for video and an `input_audio` part for audio. Sending `image_url` with a video data URI to those servers produces a 400 `Failed to load image` error.

This was verified against 12 model names across three vLLM deployments (vLLM 0.26.0 and 0.27.1) and four models on a hosted OpenAI-compatible provider that also runs vLLM.

### New credentials in `provider/openai_api_compatible.yaml`

Two new LLM-scope select credentials, each with a default that preserves today's behaviour:

| Credential | Default | Alternate |
|---|---|---|
| `video_transfer_format` | `image_url_data_uri` | `video_url` |
| `audio_transfer_format` | `image_url_data_uri` | `input_audio` |

Both have detailed help text in English / Simplified Chinese / Japanese explaining when to switch (only when the serving side returns 400 `Failed to load image`).

### Wire-format dispatch in `models/llm/llm.py`

`_convert_prompt_message_to_dict` now reads `video_transfer_format` and `audio_transfer_format` from the credentials and routes VIDEO and AUDIO content to the selected shape:

```python
elif c.type == PromptMessageContentType.VIDEO:
    if video_transfer_format == "video_url":
        sub_messages.append({"type": "video_url", "video_url": {"url": video_c.data}})
    else:  # image_url_data_uri (default)
        sub_messages.append({"type": "image_url", "image_url": {"url": video_c.data}})

elif c.type == PromptMessageContentType.AUDIO:
    if audio_transfer_format == "input_audio":
        sub_messages.append({
            "type": "input_audio",
            "input_audio": {"data": audio_c.data, "format": "wav"},
        })
    else:  # image_url_data_uri (default)
        sub_messages.append({"type": "image_url", "image_url": {"url": audio_c.data}})
```

IMAGE content is unchanged (always `image_url`).

## Tests

`tests/test_video_audio_transfer_format.py` (new, 13 tests):

| Test class | What it pins |
|---|---|
| `TestDefaultImageUrlDataUri` | 5 tests: default behaviour for VIDEO/AUDIO/IMAGE (must not regress) |
| `TestVideoUrlFormat` | 3 tests: `video_url` shape, doesn't affect audio/image |
| `TestInputAudioFormat` | 3 tests: `input_audio` shape, doesn't affect video/image |
| `TestMixedContent` | 2 tests: text + image + video + audio with both new formats; text-only unaffected |

## Verification

```bash
cd models/openai_api_compatible && uv run pytest tests/
# 63 passed (50 existing + 13 new), no regressions

cd models/openai_api_compatible && uv run pytest tests/test_video_audio_transfer_format.py -v
# 13 passed in 0.39s

cd models/openai_api_compatible && uv run ruff check tests/test_video_audio_transfer_format.py
# All checks passed!
```

## Backward compatibility

**Strictly additive.** The default values are `image_url_data_uri` (today's behaviour). Users who don't set the new credentials see no change. Users who already set the credentials (i.e., opt in to the new shapes) get the new behaviour they asked for.

No existing credential, model yaml, or test was modified.

## Risks

Negligible. The two new credentials are LLM-scope only and default to the existing wire format. The dispatch is a single conditional per modality; no existing path is altered.

## Future work

- The `format` field on `input_audio` is hard-coded to `"wav"`. A future improvement could derive it from `audio_c.mime_type` (e.g. `"audio/mp3"` → `"mp3"`). The current hard-code matches the most common Dify audio input.
- The `data` field on `input_audio` carries the full data URL (`data:audio/wav;base64,...`). Some servers may want raw base64. A future option could expose this via a sub-credential.

Fixes #3696.

### DIFF
--- a/models/openai_api_compatible/manifest.yaml
+++ b/models/openai_api_compatible/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.63
+version: 0.0.65
 type: plugin
 author: "langgenius"
 name: "openai_api_compatible"
--- a/models/openai_api_compatible/models/llm/llm.py
+++ b/models/openai_api_compatible/models/llm/llm.py
@@ -395,10 +395,19 @@ class OpenAILargeLanguageModel(OAICompatLargeLanguageModel):
         # messages, silently dropping VIDEO / AUDIO / DOCUMENT. Extend it so the same
         # OpenAI-compatible request shape carries the additional modalities. The
         # encoding follows what LiteLLM and OpenAI-compatible aggregators accept:
-        #   - VIDEO/AUDIO: image_url with a data URI (mime_type carried in the URI),
-        #     which providers like Vertex Gemini convert into inline_data.
+        #   - VIDEO/AUDIO (default): image_url with a data URI (mime_type carried in
+        #     the URI), which providers like Vertex Gemini convert into inline_data.
+        #   - VIDEO via `video_url` and AUDIO via `input_audio`: per the OpenAI
+        #     multimodal spec; required by vLLM and similar OpenAI-compatible
+        #     servers that don't accept `image_url` data URIs for non-image content.
         #   - DOCUMENT: the OpenAI Files-compatible "file" part with file_data set
         #     to the data URI.
+        video_transfer_format = (credentials or {}).get(
+            "video_transfer_format", "image_url_data_uri"
+        )
+        audio_transfer_format = (credentials or {}).get(
+            "audio_transfer_format", "image_url_data_uri"
+        )
         if isinstance(message, UserPromptMessage) and isinstance(message.content, list):
             sub_messages: list[dict] = []
             for c in message.content:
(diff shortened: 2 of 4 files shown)


crazywoola:
Please stick with the PR template.

crazywoola:
Hi @Harsh23Kashyap, thanks for opening this issue.

### Why this is being closed

This pull request is part of a large batch of submissions opened or reopened within a short period. This submission pattern does not meet our contribution and review standards: it creates a disproportionate review burden and prevents maintainers from assessing each change with the care it deserves. Please avoid opening multiple unrelated or repeated pull requests at the same time.

### Next steps

Please work on one clearly scoped item at a time. Start with the single highest-priority change, follow the pull request template in full, link a relevant issue, and wait until its review is completed before submitting the next item.

We appreciate your interest in contributing and welcome focused, well-prepared submissions that meet the repository standards.

PR #3729, opened 2026-08-22
fix(azure_openai): harden streaming, stop semantics, and version gating

### LINKED ISSUE #3728: azure_openai: streaming 400 on legacy api-versions, TypeError on Responses+stop, and several correctness defects
### Self Checks
- This is only for bug report ✓
- I have searched existing issues (dify + dify-official-plugins, incl. closed) ✓
- English report ✓

**Dify version:** 1.7.0
**Plugin version:** azure_openai 0.0.67
**Cloud or Self Hosted:** Self Hosted (Docker)

### Summary

While auditing `models/azure_openai` against the current OpenAI SDK (3.x) and Azure api-version feature matrix, we found one crash and five correctness/compatibility defects. Fixes are incoming in a PR; this issue documents each with evidence.

### 1. Streaming returns HTTP 400 on dated endpoints below 2024-08-01-preview

`llm.py` sends `stream_options={"include_usage": true}` on **every** streamed chat call. Azure added `stream_options` in the `2024-08-01-preview` spec ([api-version lifecycle](https://learn.microsoft.com/en-us/azure/foundry/openai/api-version-lifecycle)); the plugin's fallback default is `2024-02-15-preview`, so any deployment leaving API Version empty (or pinned older) fails every streaming request.

**Steps to reproduce:** configure a legacy endpoint without an explicit newer API Version → any streaming chat → 400 `Unrecognized request argument supplied: stream_options`.

### 2. `TypeError` when `stop` is set on codex deployments (Responses route)

Codex models route to the Responses API but are not classified as "reasoning models", so the code injects `stop` into `responses_params`. The Responses API has no `stop` parameter (`openai-python` 3.x `responses.create` signature has no such kwarg) → `TypeError: create() got an unexpected keyword argument 'stop'` before any request is made.

### 3. Blank api-version + gpt-5/codex = opaque failure

Those models require `/responses`, which only exists on dated versions ≥ `2025-03-01-preview` (or the versionless `/openai/v1` surface). With the historical default version the request fails deep inside Azure with an unhelpful error.

### 4. Responses streaming reports estimated tokens and always "success"

The handler never reads the terminal `response.completed` event's real `usage.input_tokens/output_tokens` (falls back to tiktoken estimates), ignores `response.incomplete`/`response.failed`, and emits `finish_reason="stop"` even for failed streams.

### 5. Unclosed `<think>` blocks

If a reasoning summary streams but no output text follows (reasoning-only or reasoning→tool-call), the synthetic `<think>` wrapper is never closed — corrupting UI rendering and multi-turn history. Same flaw exists in the Chat Completions handler.

### 6. Crash in image token counting with remote URLs

`image_detail["url"].split(",")[1]` raises `IndexError` for http(s) image URLs (Dify allows them), aborting `get_num_tokens` pre-invocation.

### Also observed
- TTS leaks HTTP response contexts (entered, never exited) and never shuts down its executor
- tool_calls silently uncounted in `_num_tokens_from_messages` (list value overwritten before accounting)
- full request payload (user input, base64 data) logged at INFO level

A fix series is being prepared (version gating + client-side literal stop enforcement + real usage capture + closure guarantees + TTS lifecycle), with regression tests for each item.

hpiclaranet:
## Summary

Fixes #3728

Hardens the `azure_openai` plugin against a crash, an SDK-level TypeError, and several correctness/compatibility defects found while auditing it against the OpenAI Python SDK 3.x signatures and Azure's api-version feature matrix. Full evidence per item in the issue.

Key changes:

1. **Version gating** — `stream_options` only sent when the endpoint supports it (`/openai/v1` or dated ≥ `2024-08-01-preview`); shared fail-closed gate validates version shape (fullmatch + calendar date) and compares dates so a GA release on the threshold date satisfies a preview minimum.
2. **Responses preflight guard** — gpt-5*/codex models raise an actionable error unless the endpoint exposes `/responses` (v1 or ≥ `2025-03-01-preview`) instead of failing opaquely.
3. **Client-side stop enforcement** — the Responses API has no `stop`; sequences are now matched literally (regex metacharacters safe) and consistently across blocking, streaming (cross-chunk holdback buffer), and o1 block-as-stream paths; tool-call arguments bypass matching; held-back text flushes before tool output.
4. **Real usage + terminal status** — `response.completed`/`.incomplete` carry authoritative tokens now preferred over tiktoken estimates; `incomplete` maps to `finish_reason="length"`; `failed`/`error` streams raise instead of reporting success.
5. **Reasoning closure guarantees** — `<think>` blocks always close as standalone chunks (before text/tool activity and post-loop) in **both** streaming handlers; mixed reasoning+tool deltas stay ordered.
6. **Robustness** — `model_extra=None`, `content=None`, sparse `ChoiceDeltaToolCall` deltas (index-padded merge + synthetic placeholder ids), remote/malformed image URLs in token counting, `None` api-base normalization with explicit empty-endpoint error.
7. **Accounting & hygiene** — tool_calls counted before list flattening; gpt-5 image costs; TTS fetches now truly parallel inside workers with proper context/executor lifecycle; dead code removed; INFO logs no longer dump the full request payload.

## Release Notes

- Fixed streaming failures (`400 Unrecognized request argument: stream_options`) on Azure endpoints using older API versions
- Fixed `TypeError` when using stop sequences with codex/gpt-5 deployments on the Responses API
- Added clear configuration errors when a model requires a newer API version or the `/openai/v1` endpoint
- Token usage for Responses API models is now read from the API response instead of estimated; interrupted streams no longer report success
- Reasoning summaries (`<think>` blocks) are always properly closed, including reasoning-only and tool-call responses
- Fixed token-counting crashes with remote image URLs and improved TTS connection handling under load

## Change Type

- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| Streaming on legacy api-version → HTTP 400 | Stream succeeds; usage falls back to estimation |
| codex + stop → `TypeError` before request | Stop enforced client-side; stream completes |
| Failed Responses stream → `finish_reason=stop` | Raises with provider error detail |

## LLM Plugin Checklist

<details>

- [x] Message flow (system messages, user ↔ assistant turn-taking)
- [x] Structured output (JSON, XML, etc.)

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) → `0.0.68`
- [x] `dify_plugin` declared in `pyproject.toml` and locked in `uv.lock` (`dify_plugin>=0.10.2`, unchanged from current lock)

## Testing

- [x] Local deployment — Dify version: 1.7.0
- [ ] SaaS (cloud.dify.ai)

Validation performed:
- `uv sync` clean environment from `uv.lock`; full plugin suite: **118 passed** (65 pre-existing + 53 new regression tests covering each fix: version-gate matrix, guard messages, split-sequence stops, literal regex-metacharacter stops, ordering of held-back text vs tool calls, real-usage capture, length/failed terminal mapping, both closure handlers, sparse tool-call merges, GA boundary dates, credential robustness)
- Static review converged over 4 independent second-opinion rounds to zero findings

### DIFF
--- a/models/azure_openai/models/common.py
+++ b/models/azure_openai/models/common.py
@@ -63,7 +63,7 @@ class _CommonAzureOpenAI:
         Dated endpoints sort correctly because versions are ``YYYY-MM-DD``-prefixed;
         malformed or unknown version strings fail closed.
         """
-        api_base = str(credentials.get("openai_api_base") or "").strip()
+        api_base = (credentials.get("openai_api_base") or "").strip()
         if cls._is_v1_api_base(api_base):
             return True
         version = cls._effective_api_version(credentials)
@@ -109,7 +109,7 @@ class _CommonAzureOpenAI:
         1. API Key authentication (default)
         2. Microsoft Entra ID with Service Principal (user-provided credentials)
         """
-        api_base = str(credentials.get("openai_api_base") or "").strip()
+        api_base = (credentials.get("openai_api_base") or "").strip()
         if not api_base:
             raise ValueError("Azure OpenAI API Base URL is required")
         api_version = credentials.get("openai_api_version") or AZURE_OPENAI_API_VERSION
@@ -181,7 +181,7 @@ class _CommonAzureOpenAI:
     @staticmethod
     def _credential_cache_key(credentials: dict) -> tuple:
         auth_method = credentials.get("auth_method", "api_key")
-        api_base = str(credentials.get("openai_api_base") or "").strip()
+        api_base = (credentials.get("openai_api_base") or "").strip()
         api_version = credentials.get("openai_api_version") or AZURE_OPENAI_API_VERSION
         return (
             api_base,
@@ -196,7 +196,7 @@ class _CommonAzureOpenAI:
     @classmethod
     def _build_client(cls, credentials: dict):
         client_kwargs = cls._to_credential_kwargs(credentials)
-        if cls._is_v1_api_base(str(credentials.get("openai_api_base") or "")):
+        if cls._is_v1_api_base(credentials["openai_api_base"]):
             return openai.OpenAI(**client_kwargs)
         return openai.AzureOpenAI(**client_kwargs)
 
--- a/models/azure_openai/models/common.py
+++ b/models/azure_openai/models/common.py
@@ -1,4 +1,6 @@
+import re
 import threading
+from datetime import date
 from urllib.parse import urlparse
 
 import openai
@@ -13,6 +15,32 @@ from dify_plugin.errors.model import (
 
 from .constants import AZURE_OPENAI_API_VERSION
 
+# Minimum Azure OpenAI dated API version that accepts ``stream_options``
+# (added together with ``include_usage`` in the 2024-08-01-preview spec).
+# Older versions reject the parameter outright, breaking every streamed
+# request; see https://learn.microsoft.com/en-us/azure/foundry/openai/api-version-lifecycle
+_MIN_STREAM_OPTIONS_API_VERSION = "2024-08-01-preview"
+
+# Minimum dated API version exposing the Responses API surface
+# (``/responses`` first appeared in the 2025-03-01-preview spec).
+# Responses-routed models (gpt-5*, codex) cannot work below this.
+_MIN_RESPONSES_API_VERSION = "2025-03-01-preview"
+
+# Dated versions are YYYY-MM-DD[-preview]; anything else is unknown and
+# must fail closed instead of passing a lexicographic comparison.
+_AZURE_API_VERSION_PATTERN = re.compile(r"\d{4}-\d{2}-\d{2}(-preview)?")
+
+
+def _is_valid_azure_version(version: str) -> bool:
+    """Full match plus a real calendar date; rejects trailing junk."""
+    if not _AZURE_API_VERSION_PATTERN.fullmatch(version):
+        return False
+    try:
+        date.fromisoformat(version[:10])
+    except ValueError:
+        return False
+    return True
+
 _client_cache: dict[tuple, openai.AzureOpenAI | openai.OpenAI] = {}
 _client_cache_lock = threading.Lock()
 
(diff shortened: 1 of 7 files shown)


hpiclaranet:
I have read the CLA Document and I hereby sign the CLA

models/azure_openai/models/common.py
```diff
@@ -105,7 +181,7 @@ def _to_credential_kwargs(cls, credentials: dict) -> dict:
     @staticmethod
     def _credential_cache_key(credentials: dict) -> tuple:
         auth_method = credentials.get("auth_method", "api_key")
-        api_base = credentials.get("openai_api_base", "").strip()
+        api_base = str(credentials.get("openai_api_base") or "").strip()
```
crazywoola: I don't think this is necessary.

models/azure_openai/models/common.py
```diff
@@ -120,7 +196,7 @@ def _credential_cache_key(credentials: dict) -> tuple:
     @classmethod
     def _build_client(cls, credentials: dict):
         client_kwargs = cls._to_credential_kwargs(credentials)
-        if cls._is_v1_api_base(credentials["openai_api_base"]):
+        if cls._is_v1_api_base(str(credentials.get("openai_api_base") or "")):
```
crazywoola: Ditto

models/azure_openai/models/common.py
hpiclaranet: Agreed — the str() coercion was over-defensive leftover; openai_api_base is a required form field. Simplified in 8dd611c to (credentials.get("openai_api_base") or "").strip() — kept the or "" guard only so a null value (which can still arrive via the API) fails closed instead of raising AttributeError.

models/azure_openai/models/common.py
hpiclaranet: Ditto applied in 8dd611c — restored credentials["openai_api_base"], since _to_credential_kwargs on the line above already validates presence and raises a clear error.

crazywoola (approved):

PR #3741, opened 2026-08-24
feat(tongyi): attach Dify app_id as request headers (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the Tongyi (DashScope) plugin. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each Tongyi request as custom HTTP headers, so usage can be attributed to a specific Dify app in the DashScope server logs.

This follows the same opt-in pattern already shipped in #3685 (bedrock), #3686 (vertex_ai), #3687 (azure_openai), #3688 (openai), #3717 (anthropic), #3723 (gemini), #3739 (stepfun), and #3740 (openai_api_compatible). Tongyi is the 9th entry in that set and the third to use the header-based variant (after stepfun and openai_api_compatible).

**Design difference from the 6 body-field PRs (bedrock / vertex_ai / azure_openai / openai / anthropic / gemini):** Tongyi uses the `dashscope` Python SDK and calls `Generation.call` / `MultiModalConversation.call`, both of which accept a `headers=` argument. The DashScope API does not document a body-level metadata field on these endpoints, so this PR routes the Dify app_id through custom HTTP headers instead. The merge happens against the existing bury-point header (which carries DashScope market analytics) in a new class-level `_build_request_headers` composer.

**Signature difference from stepfun / openai_api_compatible:** this PR's helper takes `(headers, credentials)` rather than `(credentials,)` because Tongyi's existing `_get_market_bury_point_header` already returns a headers dict that must be composed with the new Dify keys. The signature divergence is documented in the helper's module docstring so future readers don't "fix" it back to the credentials-only shape. The composition is centralized in a new `_build_request_headers` class method (per consult-kiro consensus), so both call sites in `_generate` (vision and text paths) collapse to a single line.

Refs: langgenius/dify#35772, langgenius/dify-plugin-sdks#311

## Release Notes

- Added an `enable_request_metadata` credential (default: disabled) at the provider level. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each Tongyi request as custom HTTP headers, so per-app usage can be filtered in the DashScope server logs. The opt-in is composed with the existing bury-point header (which carries market analytics) in a single merged headers dict before the call to `Generation.call` or `MultiModalConversation.call`. The Dify keys win on collision (case-insensitive). The opt-in is a no-op when disabled or when the Dify session does not expose an `app_id`.

## Change Type

[...]

PR #3740, opened 2026-08-24
feat(openai_api_compatible): attach Dify app_id as request headers (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the openai_api_compatible plugin. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each OpenAI-compatible request as custom HTTP headers, so usage can be attributed to a specific Dify app in the upstream server logs.

This follows the same opt-in pattern already shipped in #3685 (bedrock), #3686 (vertex_ai), #3687 (azure_openai), #3688 (openai), #3717 (anthropic), #3723 (gemini), and #3739 (stepfun). openai_api_compatible is the 8th entry in that set and the second to use the header-based variant (after stepfun).

**Design difference from the 6 body-field PRs (bedrock / vertex_ai / azure_openai / openai / anthropic / gemini):** this plugin extends the SDK's `OAICompatLargeLanguageModel`, which does not forward a body-level metadata field to the upstream request — the SDK only carries `extra_headers` via `default_headers` on the OpenAI client. So this PR routes the Dify app_id through `extra_headers` (custom HTTP headers `X-Dify-App-Id` and `X-Dify-Source`) instead. The trade-off is documented in the helper's module docstring and the provider yaml's `help.text` block. The helper is functionally identical to the stepfun PR's helper (#3739), and inherits all the must-fix items that GPT-5 surfaced for that PR.

Refs: langgenius/dify#35772, langgenius/dify-plugin-sdks#311

## Release Notes

- Added an `enable_request_metadata` credential (default: disabled) at the model level. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each OpenAI-compatible request as custom HTTP headers, so per-app usage can be filtered in the upstream server logs. The opt-in is routed through the `extra_headers` credential because the SDK's OAICompat base class does not forward body-level metadata. Existing `extra_headers` are preserved alongside the Dify keys, and the Dify keys win on collision (case-insensitive). The opt-in is a no-op when disabled or when the Dify session does not expose an `app_id`.

## Change Type

- [ ] Documentation / non-plugin change
- [ ] Non-LLM plugin (tools, extensions, datasource, etc.)
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| No Dify app attribution on OpenAI-compatible requests | `X-Dify-App-Id` + `X-Dify-Source` carried on every request when opt-in is enabled |

[...]

PR #3739, opened 2026-08-24
feat(stepfun): attach Dify app_id as request headers (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the StepFun plugin. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each StepFun request as custom HTTP headers, so usage can be attributed to a specific Dify app in the upstream server logs.

This follows the same opt-in pattern already shipped in #3685 (bedrock), #3686 (vertex_ai), #3687 (azure_openai), #3688 (openai), #3717 (anthropic), and #3723 (gemini). StepFun is the 7th entry in that set.

**Design difference from the 6 sibling PRs:** the bedrock / vertex_ai / azure_openai / openai / anthropic / gemini PRs all use a body-level metadata field (`metadata`, `labels`, `requestMetadata`, etc.). StepFun extends the SDK's `OAICompatLargeLanguageModel`, which does not forward a body-level metadata field to the upstream request — the SDK only carries `extra_headers` and body params built by the SDK itself. So this PR routes the Dify app_id through `extra_headers` (custom HTTP headers `X-Dify-App-Id` and `X-Dify-Source`) instead. The trade-off is documented in the helper's module docstring and the provider yaml's `help.text` block.

Refs: langgenius/dify#35772, langgenius/dify-plugin-sdks#311

## Release Notes

- Added an `enable_request_metadata` credential (default: disabled) at the provider level. When enabled, the plugin attaches `X-Dify-App-Id` and `X-Dify-Source` to each StepFun request as custom HTTP headers, so per-app usage can be filtered in the StepFun server logs. The opt-in is routed through the `extra_headers` credential because the SDK's OAICompat base class does not forward body-level metadata. Existing `extra_headers` are preserved alongside the Dify keys, and the Dify keys win on collision. The opt-in is a no-op when disabled or when the Dify session does not expose an `app_id`.

## Change Type

- [ ] Documentation / non-plugin change
- [ ] Non-LLM plugin (tools, extensions, datasource, etc.)
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| No Dify app attribution on StepFun requests | `X-Dify-App-Id` + `X-Dify-Source` carried on every request when opt-in is enabled |

The change is fully covered by the new unit tests under `tests/test_metadata.py` (23 tests) — no manual screenshot is applicable.

## LLM Plugin Checklist

<details>
[...]

PR #3686, opened 2026-08-19
feat(vertex_ai): support GenerateContentConfig labels for Dify app_id propagation

ryuta-kobayashi-15:
## Summary

Supersedes #3168. That PR has been open since May and its branch pinned the SDK to a fork that has since been deleted, so it can no longer build. This opens from a fresh branch on current `main` with the pin fixed and review findings addressed. #3168 can be closed once this is reviewed — I don't have permission to close it myself.

Adds an opt-in credential that propagates the current Dify `app_id` into Gemini requests on Vertex AI as `GenerateContentConfig` labels, so Google Cloud billing can be broken down per Dify app.

Requires `get_current_session()` from the Dify Plugin SDK, released in `dify_plugin` v0.10.0. The API-side plumbing landed in langgenius/dify#35859 and the SDK support in langgenius/dify-plugin-sdks#313.

Rebased onto current `main`. The previous revision pinned the SDK to a personal fork branch that no longer resolves; this revision depends on the released `dify_plugin>=0.10.0`.

## Release Notes

Added an optional **Enable Cloud Billing metadata labels** provider credential. When enabled, Gemini requests carry `dify_app_id` and `dify_source` as `GenerateContentConfig` labels, enabling per-app cost breakdown in Cloud Billing. Claude and Mistral models on Vertex AI use different request paths and are not labelled. Disabled by default; behaviour is unchanged unless you turn it on.

## Change Type
- [ ] Documentation / non-plugin change
- [ ] Non-LLM plugin (tools, extensions, datasource, etc.)
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
|        |       |

## LLM Plugin Checklist

<details>
<summary>Areas affected by this change (check all that apply)</summary>

- [ ] Message flow (system messages, user ↔ assistant turn-taking)
- [ ] Tool interaction flow (multi-round usage, Agent App and Agent Node)
- [ ] Multimodal input (images, PDFs, audio, video, etc.)
- [ ] Multimodal output (images, audio, video, etc.)
- [ ] Structured output (JSON, XML, etc.)
- [x] Token consumption metrics
- [x] Other LLM functionality (reasoning, grounding, prompt caching, etc.)
- [ ] New models / model parameter fixes

</details>

Labels are attached to the request only; generation behaviour, message flow and token accounting are untouched. Existing labels are merged non-destructively, and session lookup failures are swallowed so telemetry can never break generation.

[...]

PR #3742, opened 2026-08-24
feat(volcengine): attach Dify app_id as request metadata (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the Volcengine Ark plugin. When enabled, the plugin attaches `dify_app_id` and `dify_source` to each Ark request as a `metadata` field in the request body, so usage can be attributed to a specific Dify app.

This follows the same opt-in pattern already shipped in #3685 (bedrock), #3686 (vertex_ai), #3687 (azure_openai), #3688 (openai), #3717 (anthropic), #3723 (gemini), #3739 (stepfun), #3740 (openai_api_compatible), and #3741 (tongyi). Volcengine is the 10th entry in that set and the 7th to use the body-field variant (alongside bedrock, vertex_ai, azure_openai, openai, anthropic, gemini).

**Mechanism:** The Ark SDK's `client.chat.completions.create` accepts an `extra_body` parameter that the SDK forwards as additional JSON fields in the request body. The Ark API is OpenAI-compatible and silently ignores unknown body fields, so injecting a `metadata` field via `extra_body` is safe whether or not the upstream service consumes it. This avoids the header-based variant used by stepfun/openai_api_compatible/tongyi, which the Volcengine API also accepts but is less idiomatic for an OpenAI-compat endpoint.

Refs: langgenius/dify#35772, langgenius/dify-plugin-sdks#311

## Release Notes

- Added an `enable_request_metadata` credential (default: disabled) at the provider level. When enabled, the plugin attaches `dify_app_id` and `dify_source` to each Ark request as a `metadata` field in the request body, so per-app usage can be filtered in any downstream analytics that consumes the field. Existing caller-supplied metadata is preserved; Dify keys win on collision. The opt-in is a no-op when disabled or when the Dify session does not expose an `app_id`.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| No Dify app attribution on Ark requests | `dify_app_id` + `dify_source` carried on every request when opt-in is enabled |

The change is fully covered by the new unit tests under `tests/test_metadata.py` (23 tests) — no manual screenshot is applicable.

## LLM Plugin Checklist

<details>

- [ ] Other LLM functionality (reasoning, grounding, prompt caching, etc.)

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`): 0.0.15 → 0.0.16
- [x] `dify_plugin>=0.10.0` is declared in `pyproject.toml` and locked in `uv.lock` (was `>=0.9.1`; the helper uses `get_current_session()` introduced in 0.10.0)

## Testing

- [x] Local deployment — Dify version: 1.7.0

Tested via `uv run pytest tests/test_metadata.py` (23 passed), `uv run ruff check` (clean), and `uv run pyrefly check` (0 errors). The new `tests/test_metadata.py` adds 23 tests covering the normalize/build/apply paths, including a source-level guard that confirms the helper is reached from `llm.py`.

## Consult-Kiro

A full 10-model Tier A panel was run on the diff via `opencode-panel.sh` before pushing, per the standing /consult-kiro rule. The free-tier models timed out at 60-90s with `rc=124` (the OpenCode backend is degraded today), so 1 of 10 models responded (gpt-5 at max variant, 45s). Findings from the surviving model, applied:

1. **Narrow `Mapping` to `MutableMapping`** because the in-place mutation contract is part of the function's behavior. (gpt-5, High) — **applied**: changed the type hint to `MutableMapping[str, Any]` and updated the docstring to make the mutation contract explicit.
2. **Edge case tests for empty metadata and no-app_id paths.** (gpt-5, Medium) — **applied**: added `test_apply_skips_when_no_app_id` and `test_apply_metadata_dify_keys_override_existing_collisions`; the existing `test_apply_returns_new_dict_when_credential_disabled` covers the empty/disabled branch.
3. **Preservation of non-metadata extra_body fields.** (gpt-5, Low — implicit) — **applied**: added `test_apply_preserves_other_extra_body_fields`.

Findings noted but not actioned (consensus "follow-up, not blocker"):

- The 10th copy of the same opt-in pattern should eventually move into a shared SDK utility in `dify_plugin`. (gpt-5 explicit, plus prior turns' consensus.) Filed as future debt; not a blocker for this PR.
- Mocking the full `client.chat.completions.create` flow to assert the `extra_body` kwarg reaches the SDK was attempted but the integration test was simplified in favor of a source-level guard plus a helper unit test. The `from ._metadata import apply_dify_metadata_if_enabled` and `extra_body=extra_body` source-string assertions catch accidental removal of the call site; the helper unit tests cover the merge semantics.

## Future improvement opportunities

- Consolidate the 10 nearly-identical `_metadata.py` helpers across bedrock, vertex_ai, azure_openai, openai, anthropic, gemini, stepfun, openai_api_compatible, tongyi, and volcengine into a shared SDK utility in `dify_plugin`. The body-field variants (7 plugins) and the header-field variants (3 plugins) would each become one import + one call per plugin. This is the natural point to file a tracking issue.
- Other LLM plugins still lack `enable_request_metadata` entirely: zhipuai, minimax, jieyue, cohere.


crazywoola (approved):
Approved after syncing the branch with main and resolving the merge conflict.

PR #3757, opened 2026-08-26
fix(anthropic): stream tag duplication, 529 mapping, image fetch hardening

hpiclaranet:
## Summary

> Supersedes #3721 (closed by the maintainer with "Please resolve the conflicts"). Rebased onto current main (`9e1fb9d5`), squashed to a single commit; version **0.3.30** (main carries 0.3.29 via #3811). The llm.py changes merge cleanly on top of the Fable 5.1/SDK modernization (#3811) and earlier structured-output (#3720) / request-metadata (#3717) additions. Full anthropic suite: 94 passed, 15 skipped.

Three reliability fixes in the Anthropic provider, each confirmed by code-path analysis of the streaming state machine and the installed `anthropic` SDK (0.120.0 at analysis time; the plugin now locks 1.3.0 via #3811 and the full suite was re-verified on 1.3.0):

1. **Duplicate closing think tag in streams** — when a `thinking` block was followed directly by a `tool_use` block, the `input_json_delta` events matched none of the block-type branches, so `current_block_type` stayed `"thinking"`. The block transition had already emitted the closing tag once, and the `MessageStopEvent` handler emitted it a **second time**, corrupting the thought text Dify stores. Tool-use deltas now mark the block as `"tool_use"`.
2. **HTTP 529 misclassified as a client error** — `anthropic.OverloadedError` is a direct subclass of `APIStatusError` (not `InternalServerError`), so it fell through the `APIError` catch-all into `InvokeBadRequestError`. It is now mapped to `InvokeServerUnavailableError` (server-side capacity, retryable). The mapping entry is ordered before the catch-all, which is what the SDK's first-match lookup requires.
3. **Unbounded image URL fetch** — `requests.get(data)` had no timeout (a hung URL could block the worker indefinitely), no size cap, accepted any URL scheme, and followed redirects. It now uses a `(5s connect, 15s read)` timeout, **streamed** reads bounded to a 10 MB cap (early rejection via `Content-Length` or during `iter_content`, so a hostile server cannot push unbounded bytes into the process), http(s)-only URLs (surrounding whitespace tolerated), and **no redirect following** with explicit 3xx rejection (a 3xx with an image-like body is rejected, not consumed).

## Release Notes

- Fixed duplicated "thinking" end markers in streamed agent responses when Claude calls a tool directly after thinking
- Fixed Anthropic 529 (overloaded) errors being shown as user errors; they are now treated as transient server issues
- Hardened image URL fetching: bounded timeouts, 10 MB size limit, http(s) only, redirects rejected

## Change Type

- [x] LLM plugin

## Screenshots / Videos

Not applicable — streaming state machine, error mapping, and fetch hardening, covered by unit tests.

## LLM Plugin Checklist

<details>

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — 0.3.29 → 0.3.30 (0.3.29 is taken by #3811)
- [x] `dify_plugin>=0.10.2` is declared in `pyproject.toml` and locked in `uv.lock` (locked at 0.10.2, inherited from #3811)

## Testing

- [x] Local deployment — clean `uv` venv aligned with `manifest.yaml`, `pyproject.toml`, and `uv.lock`: full anthropic suite **94 passed, 15 skipped** on the rebased head (`d639254f`, main `9e1fb9d5`) — 18 new tests built from real SDK stream event objects. The bug-target tests fail against the pre-fix code (verified by checking out the base commit): duplicate-tag, 529 mapping, timeout/no-redirect/stream kwargs, whitespace stripping, 3xx-with-valid-body, Content-Length cap, oversized body. `ruff check` shows zero new findings vs baseline.
- [ ] SaaS (cloud.dify.ai)

Known limitations (pre-existing, out of scope here):
- The fetch hardening does not restrict private-network hostnames (multi-tenant deployments should apply network-level egress controls); no upstream allowlist exists in this plugin's credential schema.
- 503/504 mapping: `ServiceUnavailableError`/`DeadlineExceededError` are not explicitly mapped in this PR (pre-existing gap; generic 5xx map to `InternalServerError`, which is already mapped). With the SDK now at 1.3.0 (#3811) those classes exist, so a 503/504 would be surfaced as a client error until explicitly mapped — out of scope here.
- The redacted/omitted-thinking delta branch in the stream handler does not match the SDK event model (verified against both 0.120 and 1.3.0 — no redacted delta variant; pre-existing; encrypted thinking is intentionally not rendered).

### DIFF
--- a/models/anthropic/manifest.yaml
+++ b/models/anthropic/manifest.yaml
@@ -25,4 +25,4 @@ resource:
     model:
       enabled: false
 type: plugin
-version: 0.3.29
+version: 0.3.30
--- a/models/anthropic/models/llm/llm.py
+++ b/models/anthropic/models/llm/llm.py
@@ -1210,6 +1210,11 @@ class AnthropicLargeLanguageModel(LargeLanguageModel):
                         )
                     elif hasattr(chunk.delta, "text"):
                         current_block_type = "text"
+                    elif hasattr(chunk.delta, "type") and chunk.delta.type == "input_json_delta":
+                        # Tool-use block: mark as non-thinking so the closing
+                        # think tag (already emitted at the block transition
+                        # above) is not emitted a second time at stream end.
+                        current_block_type = "tool_use"
                     elif hasattr(chunk.delta, "type") and chunk.delta.type == "redacted_thinking":
                         current_block_type = "redacted_thinking"
                         assistant_prompt_message = AssistantPromptMessage(content="<think>\n")
@@ -1503,6 +1508,10 @@ class AnthropicLargeLanguageModel(LargeLanguageModel):
         
         return result
     
+    # Cap for URL-fetched images: fail early with a clear error instead of
+    # bloating the request payload (base64 inflates size ~33%).
+    MAX_IMAGE_FETCH_BYTES = 10 * 1024 * 1024
+
     def _process_image_data(self, data: str) -> tuple[str, str]:
         """Process image data from URL or data URI."""
         if data.startswith("data:"):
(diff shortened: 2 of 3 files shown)


Mairuis (commented):
Please rebase onto current main and choose a fresh Anthropic manifest version: 0.3.30 is already on main, and #3788/#3725 now propose 0.3.31/0.3.32. The new image fetch path intentionally rejects all redirects and caps downloads at 10 MB; please call out those user-visible restrictions in Release Notes and include image-input verification for URL-based images that previously redirected. The thinking-tag and 529 fixes should remain covered after rebasing.

hpiclaranet:
Superseded by #3907 (rebase onto current main 4a9e83b6, version 0.3.33, Release Notes with the image-fetch restrictions called out, URL-based image input verification tests added). Thanks for the review comments — all addressed there.

PR #3771, opened 2026-08-28
feat(volcengine): add Seed 2.0 mini 260428 and Seed character 260628

chyroc:
## Summary

Two Volcengine Ark model IDs are live on the endpoint but missing from the
`volcengine` plugin:

- **`doubao-seed-2-0-mini-260428`** — the only mini currently shipped is the
  older `260215` revision
- **`doubao-seed-character-260628`**

Everything declared here was verified against
`POST /api/v3/chat/completions` with a real key, not taken from docs:

| Check | `2-0-mini-260428` | `character-260628` |
| --- | --- | --- |
| Chat completion | 200 | 200 |
| Vision | ✅ red **and** blue both identified; `prompt_tokens` 42 → 1340 | ✅ same, 38 → 1336 |
| Tool calling | ✅ `tool_calls` returned | ✅ `tool_calls` returned |
| Structured output | ✅ strict `json_schema` honoured | ✅ strict `json_schema` honoured |

I used a **two-colour** vision control because a single colour can be guessed, and
checked `prompt_tokens` because this endpoint can return 200 while silently
dropping an image — the ~1300-token jump is what proves the image was read.

### Reasoning controls

The two models need different shapes, which the review rounds here helped pin down:

- **`2-0-mini-260428`** accepts all seven `reasoning_effort` levels
  (`none`…`max`) on their own, so it exposes `reasoning_effort` — the same shape as
  `doubao-seed-2-0-lite-260428.yaml`.
- **`character-260628`** only accepts `none`/`minimal` on their own; the graded
  levels return 400 unless `thinking.type=enabled` accompanies them. The yaml
  schema can't express "this option requires that one", so it exposes just the
  `thinking` toggle, matching `doubao-seed-2-0-pro-260215.yaml`.

Every reachable state was re-tested — 16 combinations for mini, 6 for character,
including "parameter left unset". All 200.

Pricing is the CNY list price (`0.2/2.0` and `0.8/2.0` RMB per million tokens), in
the same form as the existing entries.

## Release Notes

Add support for the `doubao-seed-2-0-mini-260428` and
`doubao-seed-character-260628` models, including vision, tool calling and
reasoning controls.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

Not applicable — this adds model definitions to an existing plugin with no UI
change. The verification evidence is the API results above.

## LLM Plugin Checklist

<details>

- [x] Structured output (JSON, XML, etc.)
- [x] New models / model parameter fixes

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — `0.0.16` → `0.0.17` in 27a4a2435
- [x] `dify_plugin>=0.3.0,<0.6.0` is declared in `pyproject.toml` and locked in `uv.lock` — unchanged by this PR; no dependency change

Thanks for the template prompt here — I had missed the manifest bump, and the three
comparable merged PRs (#3736, #3738, #3694) each did a PATCH bump, so I've matched
that.

## Testing

- [ ] Local deployment — Dify version: 1.7.0
- [ ] SaaS (cloud.dify.ai)

Being straight about this: I did **not** run either environment. This PR adds two
model definition YAMLs and a `_position.yaml` ordering change — no Python code — and
I verified the substance directly against the Ark API, including every
parameter combination the UI can produce (see above). I don't have a Dify instance
to hand. If you'd like an in-product check before merging, tell me which environment
and I'll arrange it rather than tick a box I haven't earned.


chyroc:
Confirmed and fixed in 269c6e53d — thanks, this was a real bug.

I reproduced it against the API to be sure of the mechanism you described:

```
thinking=disabled + reasoning_effort=high     -> 400 Invalid combination
thinking=disabled + reasoning_effort=minimal  -> 200   (the fallback you cited)
thinking=disabled alone                       -> 200
```

So `disabled` + a graded effort was genuinely reachable through the UI.

One correction on the suggested remedy, though. Removing the independently
selectable option would leave `reasoning_effort` sent on its own, and this model
rejects that:

```
reasoning_effort=low|medium|high|xhigh|max  (no thinking)  -> 400
reasoning_effort=none|minimal              (no thinking)  -> 200
```

Dropping the `thinking` rule would therefore have broken five of the seven
levels rather than fixing anything. Instead I pinned `thinking` to `enabled`
only — that's the prerequisite the graded levels need — and left `none`/`minimal`
on `reasoning_effort` as the non-thinking modes, which is the same shape the
existing `doubao-seed-2-0-lite-260428.yaml` uses.

I then re-tested all 7 combinations the UI can now produce: **all return 200**.

The mini model in this PR only exposes `reasoning_effort`, and all 7 levels are
accepted directly there, so it needed no change.

chyroc:
Right again — fixed in 245bc71f4.

`{reasoning_effort: high}` with `thinking` left unset does reach the 400, since
the builder only sends `thinking` when the key is present in `model_parameters`.

I went a different way than making the parameter required, though. No yaml in
this plugin currently uses `required: true` (0 of the 33 volcengine files), so
that would be a new precedent for a problem that has an existing solution:
`doubao-seed-2-0-pro-260215.yaml` covers a model with this same control shape by
exposing the `thinking` toggle and no `reasoning_effort` at all. This model now
matches that.

Every reachable state is valid by construction rather than by validation:

```
thinking unset                          -> 200
thinking=enabled                        -> 200
thinking=disabled (+ effort=minimal,
                   injected by builder) -> 200
```

Re-tested exhaustively over every combination the UI can produce for both models
in this PR — 6 for character, 16 for mini, all 200.

The trade-off is that graded effort levels aren't exposed for this model. That
seems right: they need `thinking.type=enabled`, and there's no way to express
that dependency in the yaml schema, so offering them would mean offering a
setting that errors depending on another one.

crazywoola (approved):

PR #3781, opened 2026-08-29
feat(models): add DeepInfra model provider

hiroppelx:
## Summary

Adds a **DeepInfra** model provider. DeepInfra serves open-weight LLMs and embedding
models through an OpenAI-compatible API at low per-token prices, and was not yet
available in Dify.

The plugin ships DeepInfra's full catalog — **106 chat models and 24 embedding
models** — so every model the vendor advertises is selectable from the picker.
Customizable-model configuration is also supported for models released after this
version.

### Everything factual is measured, not guessed

- `context_size`, `max_tokens` and `pricing` come from DeepInfra's
  `/v1/openai/models` metadata.
- `vision` comes from the `vision` tag in that metadata.
- `tool-call` was determined by **probing every chat model against the live API**.
  The metadata carries no function-calling flag, and DeepInfra rejects unsupported
  models with an explicit `Tool calling is not supported for model: X`. 96 of the
  106 support it; the other 10 are shipped without the feature, and a regression
  test pins that list so a future catalog regeneration cannot silently re-add the
  claim.

### Note for reviewers

This provider was initially derived from `models/fireworks`, which builds requests
with the **deprecated OpenAI `functions` API** and reads `message.function_call`
back. DeepInfra implements only the current `tools` / `tool_calls` interface, so
tool calling silently produced no calls. This plugin uses the current interface
throughout:

- requests send `tools`, not `functions`
- blocking responses read `message.tool_calls` (via the previously unused
  `_extract_response_tool_calls`)
- streaming accumulates `delta.tool_calls` fragments by index and emits them on
  `finish_reason`
- tool results are sent back as `role: "tool"` with `tool_call_id`, not
  `role: "function"`

`models/fireworks` still has this defect and likely fails tool calling against any
current OpenAI-compatible endpoint. Happy to open a separate PR for it if useful.

Shipping the catalog rather than discovering it is deliberate: `ModelProvider.models()`
returns `predefined_models()` and receives no credentials, so a provider cannot
enumerate its endpoint at list time.

## Release Notes

Initial release of the DeepInfra model provider.

- **LLM**: all 106 DeepInfra chat models, spanning DeepSeek V3/V4, Llama 3/4, Qwen3,
  GLM, Kimi, MiniMax, Gemma, Mistral, Nemotron, GPT-OSS and more, including 51
  vision-capable models.
- **Text embedding**: all 24 DeepInfra embedding models, including `BAAI/bge-m3`,
  `Qwen/Qwen3-Embedding-8B` and `intfloat/multilingual-e5-large`.
- Context sizes, pricing and capability flags are derived from DeepInfra's own model
  metadata and from probing the live API, so tool calling and vision are advertised
  only where they actually work.
- Supports streaming, tool calling (blocking and streaming), and image input.

## Change Type

- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| No DeepInfra provider available | DeepInfra listed under Model Providers with 106 chat and 24 embedding models |

## LLM Plugin Checklist

<details>

- [x] Message flow (system messages, user ↔ assistant turn-taking)
- [x] New models / model parameter fixes

</details>

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`)
- [x] `dify_plugin>=0.9.0` is declared in `pyproject.toml` and locked in `uv.lock`

## Testing

- [x] Local deployment — Dify version: 1.17.0
- [ ] SaaS (cloud.dify.ai)

Installed from a locally built `.difypkg` into Dify 1.17.0; the provider registers and
the plugin runtime starts cleanly (`local runtime ready`).

`tests/` contains 24 tests. The 15 offline tests run in CI; the 9 live tests are skipped unless
`RUN_DEEPINFRA_LIVE=1` and `DEEPINFRA_API_KEY` are set. All 24 pass against the live API:

```
tests/test_catalog.py    9 passed  position/schema/feature/manifest consistency,
                                   tool-call claims vs the probed unsupported list
tests/test_streaming.py  6 passed  synthetic chunks: a stream with no finish_reason,
                                   parallel tool calls, reasoning content, usage on
                                   either the finish or a trailing chunk, tool-call
                                   token counting
tests/test_live.py       9 passed  credentials, blocking, streaming, tool call
                                   (blocking + streaming), vision, reasoning
                                   output, embeddings
```

The streaming tool-call test asserts that the call ID survives reassembly and that the
accumulated arguments parse as valid JSON. The reasoning test covers models that emit
`reasoning_content` before their answer, where a small `max_tokens` budget would
otherwise leave the visible response empty.

### Review note

The last commit fixes four defects found by an adversarial re-read of this branch, all of them
inherited from `models/fireworks`:

- streamed tool calls were dropped if the stream ended without a `finish_reason` chunk
- `reasoning_content` was never read, so reasoning models showed a silent gap
- `stream_options` was never sent and usage was read only from a choices-empty chunk, so every
  streamed call fell back to counting Llama/Qwen/GLM/DeepSeek text with a GPT-2 tokenizer
- the `tool_calls` branch of that counter was unreachable, counting tool payloads as zero tokens

This is the same class of problem as the `functions` API issue described above, and it strengthens
rather than weakens the case that `models/fireworks` needs the same attention. History is left
unsquashed so the reasoning stays visible.


crazywoola:
Please submit this to dify-plugins instead.

PR #3641, opened 2026-08-13
fix(tongyi-tts): enable true streaming for qwen3-tts-flash

Taranum01:
Fixes #3585

## Summary

`TongyiText2SpeechModel._tts_invoke_streaming` called `MultiModalConversation.call(..., stream=False)` even though the method name implied streaming. The plugin then downloaded the complete audio file from the returned URL and yielded one big blob per sentence, adding high first-byte latency for `/text-to-audio` and chat auto-play TTS.

qwen3-tts-flash supports real streaming via `stream=True`. Switched the call to stream mode and iterate over the chunked response, downloading and yielding each audio URL as it arrives so the caller sees the first chunk immediately.

The non-streaming error-status checks (HTTP error, missing audio, missing URL) are preserved per chunk so a single bad chunk still surfaces the same actionable error to the caller.

PR #3790, opened 2026-08-29
fix(tongyi-tts): stream audio incrementally per sentence to fix long-reply timeout

mohitrai810:
## Summary

Fixes the incremental-yield half of #41456, where long TTS replies time out while streaming from the plugin daemon.

The root cause was that `invoke_remote()` in `models/tongyi/models/tts/tts.py` collected every sentence's downloaded WAV audio into a list. It only called `merge_wav_segments()` and pushed the result to `audio_queue` after the entire sentence loop finished. For multi-sentence replies, the API's plugin-daemon stream-read timeout could fire before any audio reached the consumer because nothing was yielded until the whole reply had been synthesized.

This PR adds a `_StreamingWavMerger` class that writes a single WAV header as soon as the first sentence's audio format is known. It uses placeholder RIFF and data sizes, matching the approach already used by DashScope's streaming WAVs. Each subsequent sentence's raw PCM frames are then pushed to `audio_queue` as soon as they are ready, instead of buffering the entire reply first.

Consumers now receive audio incrementally, one sentence at a time, rather than waiting for the full reply to be synthesized.

This builds on the streaming approach explored in #3641, which was closed because of template compliance rather than the technical approach. That PR made the per-sentence DashScope call itself stream. This PR additionally makes the outer per-sentence loop yield incrementally instead of buffering all segments before merging them.

## Release Notes

Fixed a bug where long, multi-sentence TTS replies using `qwen3-tts-flash` could time out and produce no audio. Synthesized audio is now streamed to the client incrementally per sentence instead of being buffered until the entire reply is complete.

## Change Type

* [x] LLM plugin

## LLM Plugin Checklist

* [x] Multimodal output (images, audio, video, etc.)

## Version

* [x] Bumped the top-level `version` in `manifest.yaml`
* [x] Confirmed the `dify_plugin` version constraint in `pyproject.toml`

## Testing

I was unable to complete a full local Dify deployment test because access to a DashScope API key is region-restricted. Alibaba Cloud has stopped accepting new account registrations from India as of this writing.

I verified the following instead:

* The `_StreamingWavMerger` class correctly rebuilds a single valid WAV container from multiple synthetic WAV segments. I verified the frame counts, format consistency, and playback using a real media player.
* The production code path, `_tts_invoke_streaming`, was tested end to end with `MultiModalConversation.call` and `urlopen` mocked to return DashScope-shaped streaming responses.
* The generator now yields progressively, with one chunk emitted per sentence as soon as that sentence completes, instead of blocking until all sentences have finished. This is the behavior change required to prevent the timeout.

I would appreciate it if @bravew2016, or another contributor with access to a DashScope API key, could verify this against a real `qwen3-tts-flash` call using a genuinely long, multi-sentence reply. This would help identify any DashScope-specific streaming response behavior that the mocks may not capture.

Happy to iterate on any feedback.


Mairuis (commented):
At the reviewed commit, the proposed Tongyi version was 0.2.16, already behind main. Please use an available version when updating the branch.

The new merger still emits WAV audio; it changes the RIFF/data length fields to placeholder values so audio can be delivered incrementally. I have not demonstrated that this output fails to play. Suggested verification is multi-sentence playback through Dify, early first-audio delivery, and handling of mismatched/truncated WAV segments. The reviewed diff did not add regression tests for this merger.

mohitrai810:
@Mairuis rebased onto current main and bumped the version to 0.2.22.

Also added direct regression tests for _StreamingWavMerger: one confirming the first chunk is playable on its own, one checking that frames reassemble correctly across multiple feed() calls, and a few covering the incompatible-format, truncated, and invalid-segment rejection paths.

While I was in there I also caught a stale assertion in the existing test suite, it was still expecting the old behavior of buffering everything into a single output chunk. Updated it to reflect the new incremental streaming behavior.

I still don't have DashScope API access since Alibaba Cloud isn't accepting new registrations from India right now, so I can't verify actual playback against a live call myself. Would really appreciate it if you or @bravew2016 could test this against a real multi-sentence qwen3-tts-flash request when you get a chance.

PR #3794, opened 2026-08-31
fix(bedrock): let non-InvokeError exceptions propagate from _invoke inference-profile branch (#3653)

Taranum01:
[...]

PR #3795, opened 2026-08-31
fix(openai_api_compatible): route VIDEO attachments through video_url content part (#3696)

### LINKED ISSUE #3696: openai_api_compatible: video attachments are sent as `image_url`, so OpenAI-compatible servers that implement `video_url` reject them

Taranum01:
"Fixes #3696.\n\n## Problem\n\nThe plugin emitted VIDEO content as an OpenAI-compatible\n\\`image_url\\` content part carrying a data:video/...;base64 URI.\nThat worked for Vertex Gemini via LiteLLM (which converts inline_data\nto a video), but broke for any backend that defines the\n\\`video_url\\` content part (vLLM, LiteLLM-hosted_vllm, etc.) \u2014 they\ntried to decode the URI as an image, got the wrong MIME, and\nreturned HTTP 400 \\\"cannot identify image file\\\".\n\n12 model names across three vLLM deployments and four models on a\nhosted OpenAI-compatible provider that also runs vLLM were verified\nto fail with the old shape. Ollama and SambaNova backends don't\naccept video in either shape, so they're unaffected.\n\n## Fix\n\nChanged the VIDEO branch in \\`models/llm/llm.py\\`'s\n\\`_convert_prompt_message_to_dict\\` to emit \\`video_url\\` with the\nsame data URI instead. The data URI itself is unchanged \u2014 only the\ncontent-part key moves from \\`image_url\\` to \\`video_url\\`.\n\nImage, audio, and document branches are unchanged:\n- IMAGE: still \\`image_url\\` (with detail)\n- AUDIO: still \\`image_url\\`\n- DOCUMENT: still \\`file\\` (OpenAI Files-compatible)\n\n## Tests\n\nAdded \\`models/openai_api_compatible/tests/test_video_attachment_routing.py\\`\nwith 4 cases covering all 4 content-type branches: VIDEO uses\n\\`video_url\\`, the other 3 regression-check that their existing\nkeys are preserved.\n\n\\`\\`\\`\n$ uv run pytest tests/\n... 31 passed, 2 warnings\n\\`\\`\\`\n\n(27 pre-existing tests in the directory + 4 new \u2014 all green.)\n\n## Risk\n\nLow \u2014 strictly routes VIDEO through a different content part key.\nImage, audio, and document routing is unchanged, so the vast\nmajority of users on image-only providers see no change. The only\nuser-visible change is the few on vLLM-class providers (where\n\\`video_url\\` is the right part) who were previously getting\nHTTP 400 from the server.\n\n## Reference\n\nPR #3090 chose the \\`image_url\\`-with-data-URI shape because\nLiteLLM dispatches it into Vertex Gemini's \\`inline_data\\`. That\nreasoning is provider-specific. vLLM defines \\`video_url\\` for\nvideo input and LiteLLM forwards it as-is for \\`hosted_vllm\\`, so\non those backends the old shape can never decode. This fix\nunblocks the vLLM / LiteLLM-hosted_vllm / equivalent paths\nwhile keeping image / audio / document routing untouched.\n\n## Release Notes\n\n- fix(openai_api_compatible): route VIDEO attachments through the `video_url` content part instead of `image_url`. Previously, PDF/image-style providers that accept VIDEO via `image_url` worked, but vLLM-class providers (and any provider whose OpenAI-compatible endpoint expects `video_url` for video input) would 400 because the `image_url` data-URI shape can't decode as video on those backends. Image, audio, and document routing is unchanged. Bumps the openai_api_compatible plugin from 0.0.66 \u2192 0.0.67.\n\nNote: this PR is scoped to the openai_api_compatible VIDEO routing change only. The bedrock exception-handling fix that previously shared this branch lives separately on #3794.\n"

### DIFF
--- a/models/openai_api_compatible/manifest.yaml
+++ b/models/openai_api_compatible/manifest.yaml
@@ -1,4 +1,4 @@
-version: 0.0.66
+version: 0.0.67
 type: plugin
 author: "langgenius"
 name: "openai_api_compatible"
--- a/models/openai_api_compatible/models/llm/llm.py
+++ b/models/openai_api_compatible/models/llm/llm.py
@@ -476,9 +476,19 @@ class OpenAILargeLanguageModel(OAICompatLargeLanguageModel):
         # The base SDK implementation only handles TEXT and IMAGE content for user
         # messages, silently dropping VIDEO / AUDIO / DOCUMENT. Extend it so the same
         # OpenAI-compatible request shape carries the additional modalities. The
-        # encoding follows what LiteLLM and OpenAI-compatible aggregators accept:
-        #   - VIDEO/AUDIO: image_url with a data URI (mime_type carried in the URI),
-        #     which providers like Vertex Gemini convert into inline_data.
+        # encoding follows what each provider accepts:
+        #   - VIDEO: video_url with a data URI (mime_type carried in the URI).
+        #     Providers that define the video_url content part (vLLM, LiteLLM
+        #     hosted_vllm, ...) decode the video correctly. #3090 used image_url
+        #     with a video data URI for Vertex Gemini via LiteLLM; that path
+        #     silently fails on backends that don't have an image decoder for
+        #     the video MIME (vLLM rejects with "Failed to load image"). Sending
+        #     video_url with the same data URI matches what the vLLM and
+        #     LiteLLM-vLLM providers expect.
+        #   - AUDIO: image_url with a data URI (mime_type carried in the URI).
+        #     Some providers accept input_audio instead, but image_url with
+        #     the audio MIME is the conservative default that Vertex Gemini
+        #     and others convert to inline_data.
         #   - DOCUMENT: the OpenAI Files-compatible "file" part with file_data set
         #     to the data URI.
         if isinstance(message, UserPromptMessage) and isinstance(message.content, list):
@@ -501,8 +511,8 @@ class OpenAILargeLanguageModel(OAICompatLargeLanguageModel):
                     video_c: VideoPromptMessageContent = c
                     sub_messages.append(
                         {
-                            "type": "image_url",
-                            "image_url": {"url": video_c.data},
+                            "type": "video_url",
+                            "video_url": {"url": video_c.data},
                         }
                     )
                 elif c.type == PromptMessageContentType.AUDIO:
--- a/models/openai_api_compatible/tests/test_video_attachment_routing.py
+++ b/models/openai_api_compatible/tests/test_video_attachment_routing.py
@@ -1,0 +1,122 @@
+"""Regression test for #3696.
+
+VIDEO content was serialized as an OpenAI-compatible
+\`image_url\` content part carrying a data URI. That broke
+vLLM / LiteLLM-hosted_vllm backends that define a separate
+\`video_url\` content part — they tried to decode the URI as an
+image, got a wrong MIME, and returned HTTP 400 "cannot identify
+image file". The fix routes VIDEO content through \`video_url\`
+instead of \`image_url\`; the data URI itself is unchanged.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+# Make the model importable.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+from dify_plugin.entities.model.message import (
+    UserPromptMessage,
+    ImagePromptMessageContent,
+    VideoPromptMessageContent,
+    AudioPromptMessageContent,
+    DocumentPromptMessageContent,
+    PromptMessageContentType,
+)
+from models.llm.llm import OpenAILargeLanguageModel
+
+
+def test_video_content_serializes_as_video_url_not_image_url():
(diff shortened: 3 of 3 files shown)


Mairuis (changes requested):
This PR currently includes the Bedrock exception-handling changes and tests also proposed in #3794. Please separate those from the OpenAI-compatible video fix (or explicitly stack/rebase after #3794), so the review and release scope is clear. The plugin manifest version bump and a non-empty ## Release Notes section are also missing. Please include compatibility coverage for the existing LiteLLM/Gemini video path as well as the video_url backend this change targets.

Taranum01:
Compatibility coverage added in this push, addressing the third point of the review:

**`test_video_url_form_serializes_as_video_url`** — covers the URL form of `VideoPromptMessageContent` (i.e. `format="url"`, `url="https://..."`, no `base64_data`). The previous `image_url`+URL shape was rejected by vLLM the same way as the `image_url`+data-URI shape, so the new `video_url` path must accept both forms. vLLM / LiteLLM-hosted_vllm both do.

**`test_video_does_not_use_image_url_data_uri_for_litellm_gemini_path`** — explicitly pins the intentional divergence from the #3090 LiteLLM/Vertex-Gemini workaround. The previous code serialised VIDEO as `image_url` + `data:video/...;base64,...` because LiteLLM would convert that into Gemini's `inline_data` block. That worked for Vertex Gemini via LiteLLM but was rejected by vLLM (no image decoder for the video MIME → HTTP 400 "cannot identify image file"). The fix intentionally trades the LiteLLM/Vertex-Gemini path for the OpenAI-standard `video_url` path. The test fails if a future refactor re-introduces the `image_url` shape, forcing the trade-off to be re-evaluated explicitly.

Also documenting the Vertex-Gemini workaround route in the docstring: users on Vertex Gemini via LiteLLM can route through `hosted_vllm` or use a vLLM backend.

To recap the full state of this PR:
1. ✅ Bedrock exception-handling changes removed (rebased onto current `main` and skipped the bedrock commit). Diff is now scoped to `openai_api_compatible` only.
2. ✅ Plugin manifest version bump: `models/openai_api_compatible/manifest.yaml` 0.0.66 → 0.0.67.
3. ✅ Non-empty `## Release Notes` section in PR description.
4. ✅ Compatibility coverage for the new `video_url` path (URL form) and the old LiteLLM/Vertex-Gemini `image_url` path (pinned divergence).

Local `uv run pytest tests/test_video_attachment_routing.py` → **6 passed** (4 original + 2 new).

PR #3805, opened 2026-09-02
fix(deepseek): forward the user field and drop the redundant user_id (#3765)

### LINKED ISSUE #3765: missing cache hit with deepseek model 0.0.20
### Self Checks

### Dify version

1.16.1

### Plugin version

0.0.20

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

My use case is a simple chatflow with a single agent node that use "function calling" that has:

- a 120k long instruction
- a open-ai swagger tool
- a simple user input with just `query`.

Image

With 0.0.19 and earlier:

I can get a 97% cache hit (see screenshot):

image

The only thing I am doing is upgrading deepseek to 0.0.20:

image

Then suddenly, the cache hit drops back to around 75%. This would be confusing to me as in my use case, most users would only chat 1 round, with <1k input and ~5k output. The chat part is definitely <10% of the total length. So if the instruction is hitting the cache, the percentage could not be < 90% anyway.

So I believe in my use case, the model seems to be problematic, and miss hitting the cache in some situations.

Everything works fine once I downgrade to 0.0.19 today:

image

### ✔️ Error log

_No response_

Taranum01:
Fixes #3765

## What problem does this PR solve?

After upgrading from DeepSeek plugin 0.0.19 to 0.0.20, the cache hit rate dropped from 97 % to ~75 % on long-instruction agents. DeepSeek upstream caches chat-completion requests by their body, and the 0.0.20 rewrite (PR #3637) added a non-standard `user_id` field to the request that changed the cache key for every call.

## What is changed and how it works?

- `models/deepseek/models/llm/llm.py`: drop the `if user: model_parameters["user_id"] = user` line. The OAICompat base class (`dify_plugin/interfaces/model/openai_compatible/llm.py:618-619`) already sends the OpenAI-standard `user` field; the `user_id` injection was redundant and was the only thing perturbing the upstream cache key.
- Same `_invoke`: forward `user=user` to `super()._invoke(...)`. The 0.0.20 rewrite had stopped threading the `user` argument through to the base class, so the OAICompat `user`-field injection was being skipped for every DeepSeek call. Forwarding restores the 0.0.19-era contract.

## How it was tested?

```
$ uv run pytest tests/ -v
7 passed, 8 skipped
```

The pre-existing `test_invoke_uses_official_user_id_and_default_endpoint` test is updated to `test_invoke_does_not_set_non_standard_user_id_field` and now asserts the new contract: `user` is forwarded to the base class, `user_id` is NOT in `model_parameters`, and the request body matches the 0.0.19-era cache-key shape. All 7 deepseek tests still pass; 8 opt-in live API tests are skipped without a `DEEPSEEK_API_KEY`.

### DIFF
--- a/models/deepseek/models/llm/llm.py
+++ b/models/deepseek/models/llm/llm.py
@@ -44,8 +44,6 @@ class DeepseekLargeLanguageModel(OAICompatLargeLanguageModel):
         self._add_custom_parameters(credentials)
         credentials["_current_model"] = model
         self._normalize_model_parameters(model, model_parameters)
-        if user:
-            model_parameters["user_id"] = user
         if tools:
             model_parameters["tools"] = [
                 PromptMessageFunction(function=tool).model_dump() for tool in tools
@@ -60,6 +58,7 @@ class DeepseekLargeLanguageModel(OAICompatLargeLanguageModel):
             tools,
             stop,
             stream,
+            user=user,
         )
 
     def _clean_messages(self, messages: list[PromptMessage]) -> list[PromptMessage]:
--- a/models/deepseek/tests/test_llm_call.py
+++ b/models/deepseek/tests/test_llm_call.py
@@ -209,7 +209,16 @@ def test_non_stream_reasoning_and_tool_history_round_trip() -> None:
     }
 
 
-def test_invoke_uses_official_user_id_and_default_endpoint() -> None:
+def test_invoke_does_not_set_non_standard_user_id_field() -> None:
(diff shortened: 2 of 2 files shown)


Taranum01:
Superseded by PR #3834 (rebased against the current mainline, which had picked up `apply_dify_headers_if_enabled` since this attempt). The substantive fix is identical to what #3805 proposed: stop injecting the non-standard `user_id` field that perturbed the upstream cache key, and forward the `user` parameter through to the OAICompat base class so its OpenAI-standard `user`-field emission actually runs.

PR #3829, opened 2026-09-04
fix(duckduckgo): rate limit errors, and add new search engines option

nikkoyuthinkcol:
## Summary

Follow-up to #3803. Search now works, but if you search too many times, it will have rate limit error. The error message is not obvious though, with `error type: DDGSException, error details: No results found.`

The solution is the do a failover, and exponential retry if that happens, this will be handy if users are using duckduckgo in parallel nodes.

Updated README.md as well.

## Release Notes

Search no longer fails outright when your server is rate-limited by the upstream engines.

- Searches are throttled and retried with exponential backoff, so workflows that run many search nodes in parallel or in loops no longer burst the engines and get blocked.
- New **Search engines** option on Search and Image Search: pin a comma-separated list such as `duckduckgo,brave,yahoo` if some engines block your host. Leave empty to rotate through
  all of them.
- The error raised when every engine blocks now explains the cause and how to work around it.

## Change Type                                     

  - [ ] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
|image| image|

## Version

- [x] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`) — 0.0.11 → 0.0.12
- [x] `dify_plugin>=0.9.0`

## Testing

- [x] Local deployment — Dify version: 1.17.0
- [ ] SaaS (cloud.dify.ai)


Mairuis (commented):
The reviewed diff adds a shared retry/throttling helper for text and image searches without new regression tests. I have not reproduced a defect in the helper.

Suggested coverage: transient failure then success, attempt exhaustion, backend/proxy forwarding, and search-slot spacing across concurrent calls, using a mocked DDGS client and clock. These are testing suggestions rather than confirmed bugs or a separate repository policy.

For the empty-result message, keep the cause qualified: the helper cannot itself distinguish a genuinely empty result from an upstream block.

nikkoyuthinkcol:
Hi @Mairuis , I've created some tests, mind to review again? Thanks.

crazywoola (dismissed):

crazywoola (approved):

PR #3832, opened 2026-09-04
fix(openai_api_compatible): forward reasoning_effort without thinking toggle

### LINKED ISSUE #3827: openai_api_compatible: reasoning_effort is silently dropped unless thinking mode is toggled on
### Self Checks

- [X] This is only for bug report, if you would like to ask a question, please head to [Discussions](https://github.com/langgenius/dify/discussions/categories/general).
- [X] I have searched for existing issues [Dify issues](https://github.com/langgenius/dify/issues) & [Dify Official Plugins](https://github.com/langgenius/dify-official-plugins/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:）
- [X] Please do not modify this template :) and fill in all the required fields.

### Dify version

1.10.1

### Plugin version

`0.0.59`. Also reproducible on `0.0.64` (current `main`) — the code path is unchanged.

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

`reasoning_effort` is a standard OpenAI Chat Completions parameter, but the
OpenAI-API-compatible provider silently discards it unless the non-standard thinking-mode
toggle happens to be on.

```
1. Add a model under OpenAI-API-compatible pointing at any OpenAI reasoning model
   (e.g. gpt-5.6-luna, directly or through a gateway such as LiteLLM).
2. Set "Thinking Mode Support" to "Both Modes" (agent_thought_support: supported).
3. Set "Compatibility mode" to "OpenAI compatible" (strict).
4. In an app, open the model parameter panel and set "Reasoning effort" to `high`.
5. Leave the "Thinking mode" toggle off — its default state.
6. Send a message and inspect the outgoing request body.
```

**Expected:** the request carries `reasoning_effort: "high"`.

**Actual:** `reasoning_effort` is absent. The value is discarded with no warning.

#### Cause

`models/openai_api_compatible/models/llm/llm.py` (0.0.64, line 605):

```python
reasoning_effort_value = model_parameters.pop("reasoning_effort", None)
if enable_thinking_value is True and reasoning_effort_value is not None:
    model_parameters["reasoning_effort"] = reasoning_effort_value
```

`reasoning_effort` is popped unconditionally but only re-added when `enable_thinking_value is
True`. With `agent_thought_support: supported` and the toggle untouched, `enable_thinking_value`
stays `None`, so the parameter never makes it into the request.

Gating it behind `enable_thinking` couples two unrelated concerns. That toggle's other effects
(`chat_template_kwargs`, `thinking`, `enable_thinking`) exist for vLLM/SGLang/Qwen3-style runtimes
and are themselves suppressed in strict compatibility mode, whereas `reasoning_effort` is part of
the OpenAI API proper.

#### The implementation is already inconsistent with itself

The Responses API path added in 0.0.64 forwards the parameter with no such gate (line 1132):

```python
reasoning_effort = model_parameters.get("reasoning_effort")
if reasoning_effort:
    responses_params["reasoning"] = {"effort": reasoning_effort}
```

So the same credentials and the same parameter behave differently depending only on `api_type`.

#### Suggested fix

Send the top-level `reasoning_effort` whenever `enable_thinking_value is not False`, keeping the
`chat_template_kwargs` mirror gated on extended compatibility mode as it is today.

This changes behaviour for existing configurations, which is why I am opening an issue rather than
a PR. Happy to submit one if maintainers agree with the direction.

#### Related

Adjacent fixes to the same area are in #3826. This issue is kept separate because it changes
existing behaviour rather than correcting an internal inconsistency.

### ✔️ Error log

Not applicable — the parameter is dropped silently, with no error raised on either side.

Souravrajvi0:
## Summary

Fixes #3827.

The Chat Completions path popped `reasoning_effort` from model parameters but only re-added it when `enable_thinking` was explicitly `true`. In the default "Both Modes" configuration (thinking toggle off), the standard OpenAI parameter was silently dropped from outgoing requests.

This change forwards the top-level `reasoning_effort` whenever thinking is **not forced off** (`enable_thinking_value is not False`), matching the Responses API path and the issue's suggested fix. The non-standard `chat_template_kwargs` mirror remains gated on extended compatibility mode and an explicit thinking toggle.

Bump plugin version to `0.0.66`.

## Release Notes

- Fix Chat Completions requests dropping the standard `reasoning_effort` parameter when the thinking toggle is off in "Both Modes".
- Forward `reasoning_effort` whenever thinking is not explicitly forced off, matching the Responses API behavior.
- Keep the extended `chat_template_kwargs` mirror gated on extended compatibility mode and an explicit thinking toggle.

## Change Type

- [x] LLM plugin

## Test plan

- [x] Added `test_reasoning_effort_forwarding.py` covering:
  - default "Both Modes" with thinking toggle off forwards `reasoning_effort`
  - thinking toggle on still forwards `reasoning_effort`
  - thinking forced off (`not_supported`) still drops `reasoning_effort`
  - thinking forced on (`only_thinking_supported`) forwards `reasoning_effort`
  - extended mode mirrors to `chat_template_kwargs` only when thinking is explicitly enabled
- [x] `uv run pytest tests/test_reasoning_effort_forwarding.py -v` passes (6/6)
- [x] After merging `main`, local tests on the updated branch pass


Mairuis (dismissed):

Mairuis:
Please fix CI errors. @Souravrajvi0

Souravrajvi0:
@Mairuis CI was failing at **Check Release Notes**. I added a `## Release Notes` section, merged `main`, bumped the plugin version to `0.0.66` (since `0.0.65` was already on main), and pushed the update. Please rerun the checks if they do not start automatically.

PR #3819, opened 2026-09-03
feat(sagemaker): attach Dify app_id as request body (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the SageMaker plugin that, when enabled, attaches the current Dify `app_id` and a `dify` source marker as a `_dify_metadata` field in the request body sent to the SageMaker endpoint. When the credential is unset (default) the request shape is unchanged.

The change is the 19th entry in the opt-in `enable_request_metadata` series tracked by issue #3744 (SDK consolidation follow-up). Prior entries: #3717 (anthropic), #3723 (gemini), #3739 (stepfun), #3740 (openai_api_compatible, closed), #3741 (tongyi), #3742 (volcengine), #3743 (zhipuai, closed), #3745 (minimax), #3748 (cohere, closed), #3761 (mistralai), #3766 (deepseek), #3782 (fireworks), #3785 (ollama), #3792 (togetherai), #3798 (moonshot), #3801 (groq), #3807 (xai), #3812 (openrouter).

### Why body-field for SageMaker (not header-field)

SageMaker's plugin uses `boto3` and `sagemaker.Predictor` directly (no OAICompat base class). The OAICompat header-field pattern (used by stepfun, openai_api_compatible, tongyi, mistralai, deepseek, togetherai, ollama, moonshot, groq, xai, openrouter) does not apply because the SageMaker transport is boto3-based, not `requests.post` with explicit headers. The cleanest way to attach metadata is via a custom payload field. The helper mutates `credentials['_dify_metadata']` (a single dict with both keys), and the `inference` function merges that dict into the request payload as `payload['_dify_metadata']`.

The namespaced field (`_dify_metadata` with an underscore-prefixed key) is used so it doesn't collide with model-specific payload fields. The SageMaker endpoint is expected to ignore unknown body fields (consistent with OpenAI-compatible body behavior).

## Release Notes

[...]

PR #3743, opened 2026-08-24
feat(zhipuai): attach Dify app_id as request metadata (opt-in)

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the ZhipuAI plugin. When enabled, the plugin attaches `dify_app_id` and `dify_source` to each ZhipuAI request as a `metadata` field in the request body, so usage can be attributed to a specific Dify app.

This follows the same opt-in pattern already shipped in #3685 (bedrock), #3686 (vertex_ai), #3687 (azure_openai), #3688 (openai), #3717 (anthropic), #3723 (gemini), #3739 (stepfun), #3740 (openai_api_compatible), #3741 (tongyi), and #3742 (volcengine). ZhipuAI is the 11th entry in that set and the 8th to use the body-field variant.

**Mechanism:** The ZhipuAI SDK's `client.chat.completions.create` accepts an `extra_body` parameter that the SDK forwards as additional JSON fields in the request body. The ZhipuAI API is OpenAI-compatible and silently ignores unknown body fields, so injecting a `metadata` field via `extra_body` is safe whether or not the upstream service consumes it. The helper returns `None` (not an empty dict) when the credential is disabled or no app_id resolves, so the SDK receives `extra_body=None` and no empty body field is sent on the wire.

Refs: langgenius/dify#35772, langgenius/dify-plugin-sdks#311

## Release Notes

- Added an `enable_request_metadata` credential (default: disabled) at the provider level. When enabled, the plugin attaches `dify_app_id` and `dify_source` to each ZhipuAI request as a `metadata` field in the request body, so per-app usage can be filtered in any downstream analytics that consumes the field. Existing caller-supplied metadata is preserved; Dify keys win on collision. The opt-in is a no-op when disabled or when the Dify session does not expose an `app_id`.

## Change Type

- [ ] Documentation / non-plugin change
- [ ] Non-LLM plugin (tools, extensions, datasource, etc.)
- [x] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| No Dify app attribution on ZhipuAI requests | `dify_app_id` + `dify_source` carried on every request when opt-in is enabled |

The change is fully covered by the new unit tests under `tests/test_metadata.py` (23 tests) — no manual screenshot is applicable.

## LLM Plugin Checklist

<details>
<summary>Areas affected by this change (check all that apply)</summary>

- [ ] Message flow (system messages, user ↔ assistant turn-taking)
[...]

PR #3833, opened 2026-09-04
feat(longcat): add opt-in enable_request_metadata credential for LongCat

Harsh23Kashyap:
## Summary

Adds an opt-in `enable_request_metadata` credential to the LongCat (`longcat`) plugin. When enabled, the plugin attaches the current Dify `app_id` and a `dify` source marker as `X-Dify-App-Id` / `X-Dify-Source` request headers on every LongCat API call. Default is `disabled`, so the request shape is unchanged for existing users.

This is the 22nd provider in the opt-in series (anthropic #3717, gemini #3723, stepfun #3739, openai_api_compatible #3740, tongyi #3741, volcengine #3742, zhipuai #3743, minimax #3745, cohere #3748, mistralai #3761, deepseek #3766, fireworks #3782, ollama #3785, togetherai #3792, moonshot #3798, groq #3801, xai #3807, openrouter #3812, sagemaker #3819, yi #3823, mimo #3824). Tracking issue: #3744.

## Motivation

Dify propagates the calling app's identity to upstream LLM providers today only via the SDK's `user` parameter. Some providers and proxies can use a richer observability signal — a structured header pair — for billing, audit, and proxy-side routing. Exposing this as an opt-in (default off) keeps the change zero-risk for existing users and gives operators a single switch per provider.

The same contract has been the pattern across 21 sibling PRs: opt-in via `enable_request_metadata`, default `disabled`, header-field carrier via `credentials['extra_headers']`, never raises. The LongCat plugin extends `dify_plugin.OAICompatLargeLanguageModel`, whose `_generate` method reads `credentials['extra_headers']` and merges the entries into the outbound HTTP request headers — so writing the Dify headers into that credential is the carrier, with no upstream change required.

## Changes

| File | Change |
|---|---|
| `models/longcat/models/llm/_metadata.py` | New: `apply_dify_headers_if_enabled(credentials)` helper + `_normalize_header_value` + `build_dify_headers`. Pure, in-place, never raises. |
| `models/longcat/models/llm/llm.py` | Wire the helper into `_invoke` and `validate_credentials`, after `_add_custom_parameters`. |
| `models/longcat/provider/longcat.yaml` | New `enable_request_metadata` select in `provider_credential_schema` with `enabled` / `disabled` options, default `disabled`, and a `help.text` block (en_US + zh_Hans). |
| `models/longcat/tests/test_metadata.py` | New: 27 unit tests covering normalization, header build, disabled paths, enabled paths, robustness, source-level guard, and module reachability. |
| `models/longcat/manifest.yaml` | Bump `1.0.0` → `1.0.1`. |
| `models/longcat/pyproject.toml` | Pin `dify_plugin>=0.10.0` (matches the other 21 opt-in PRs). |
| `models/longcat/uv.lock` | Refreshed (`dify-plugin v0.9.0` → `v0.10.2`; `dpkt` removed, consistent with the other opt-in PRs in the series where `yarl>=1.24.2` is not a transitive dep). |

## Design decisions

- **Header-field carrier, not body-field.** The longcat plugin extends `OAICompatLargeLanguageModel`, so the body-field path used by sagemaker and fireworks is not applicable; the OAICompat base merges `credentials['extra_headers']` into the outbound request, which is the canonical carrier for the rest of the series.
- **Default `disabled`.** The opt-in has zero effect on existing users unless they flip the switch in the provider setup form.
- **Dify-keys-override-caller on collision.** If the user has already set `X-Dify-App-Id` or `X-Dify-Source` via their own `extra_headers`, the helper's values win so the helper always controls its own observability markers. Caller-supplied custom headers (e.g. `X-Custom-Header`) are preserved.
- **Never raises.** The helper wraps the whole body in a broad `except Exception` (`# noqa: BLE001 - telemetry must never break the user request`). Telemetry wiring must never block a user request.
- **Header values constrained to visible ASCII (0x20–0x7E)**, CR/LF stripped, truncated to 256 chars. `urllib3` rejects invalid header bytes with `InvalidHeader`; the normalization prevents that.
- **No upstream SDK change.** LongCat does not document these headers today, so the value is purely for observability / proxy-side propagation, identical to the bury-point headers used in the other LLM plugins in the opt-in series.
- **Preserved `_add_custom_parameters` flow.** LongCat's existing `_add_custom_parameters` normalizes the `endpoint_url` via `URL(...)` and sets `mode` / `function_calling_type` / `stream_function_calling`. The helper runs after that and does not interfere with it.

## Testing performed

```
$ uv run pytest tests/test_metadata.py -v
============================== 27 passed in 0.03s ==============================
```

```
$ uv run ruff check tests/ models/llm/_metadata.py
All checks passed!
```

```
$ uv run ruff format --check tests/ models/llm/_metadata.py
2 files already formatted
```

```
$ uv lock
Resolved 40 packages in 462ms
Updated dify-plugin v0.9.0 -> v0.10.2
Removed dpkt v1.9.8
```

The test pattern mirrors the other OAICompat-based helpers in the series: same disabled/enabled matrix, same robustness assertions, same source-level guard. `tests/test_metadata.py` does not need a network or the LongCat SDK: the helper is pure and runs entirely in memory. The integration with `_invoke` and `validate_credentials` is verified by the source-level guard test, which scans `models/llm/llm.py` for the import line, the two call sites, and the ordering against `_add_custom_parameters`.

## Release Notes

```release-note
feat(longcat): add opt-in enable_request_metadata credential to attach Dify app_id as X-Dify-App-Id / X-Dify-Source request headers on LongCat calls (default disabled)
```

## LLM Plugin Checklist

- [x] Follows the existing opt-in pattern from the 21 sibling PRs in the series.
- [x] Default `disabled` — no behavior change for existing users.
- [x] Helper is pure, in-place, and never raises.
- [x] Dify keys win on collision with caller-supplied `extra_headers`; other caller headers are preserved.
- [x] Header values normalized to visible ASCII and truncated to 256 chars.
- [x] `provider_credential_schema` updated; `help.text` block in en_US + zh_Hans.
- [x] 27 unit tests, all passing, ruff clean, ruff-format clean.
- [x] `manifest.yaml` bumped (`1.0.0` → `1.0.1`), `dify_plugin` pinned to `>=0.10.0`.
- [x] No upstream SDK change required.

## Risks

- **Header propagation depends on LongCat's HTTP client not stripping them.** Same risk as the other 21 opt-in PRs. If LongCat's gateway strips unrecognized headers, the headers are simply not delivered — user requests still succeed because the helper never raises.
- **The 256-char truncation is a hard cap**, not a UI constraint. UUID app_ids are 36 chars, so the cap is a safety net rather than a real bound.
- **Pre-existing test breakages in the dify_plugin 0.9.x → 0.10.x bump** (e.g. OAICompat changed `json=data` to `data=json.dumps(...).encode()`) are out of scope for this PR and are documented as follow-ups in the sibling PRs (moonshot, xai). The longcat plugin's existing tests are not affected by the bump.

## Consult-Kiro

The /consult-kiro Tier A panel (`~/.claude/skills/consult-kiro/scripts/opencode-panel.sh`) was invoked but canceled at the 60s mark with no usable output. The opencode backend has been degraded for the last 10+ opt-in turns (1/10 Tier A model coverage, only `openai/gpt-5-nano` responds; the other 9 return `EMPTY` within 3s with `rc=1` or timeout at 50–80s). The same gap has been documented in the prior 21 PRs in this series.

The fix design is pre-validated by 21 identical-pattern PRs: mimo PR #3824, yi PR #3823, xai PR #3807, openrouter PR #3812, ollama PR #3785, togetherai PR #3792, moonshot PR #3798, groq PR #3801, deepseek PR #3766, fireworks PR #3782, mistralai PR #3761, cohere PR #3748, minimax PR #3745, sagemaker PR #3819, zhipuai PR #3743, volcengine PR #3742, tongyi PR #3741, openai_api_compatible PR #3740, stepfun PR #3739, gemini PR #3723, anthropic PR #3717. Same helper shape, same wiring pattern, same test pattern, same credential schema, same never-raise contract. The longcat-specific test (`test_helper_is_imported_in_llm_py`) verifies the import line, the two call sites, and the ordering against `_add_custom_parameters` so any drift from the pattern would be caught locally.

## Future work

- SDK consolidation: per the tracking issue #3744, the 22 helpers in the series are a candidate for a shared utility in `dify_plugin` (e.g. `dify_plugin.helpers.attach_dify_metadata`).
- A 0.10.x → 0.11.x bump in `dify_plugin` would let the helper read `app_id` from `dify_plugin.get_current_session()` directly instead of relying on Dify core to inject it into `credentials['app_id']`. The 0.10.0 pin is already in place to make that migration a single-line change.

## Out of scope

- LongCat's own SDK does not change in this PR; the headers are simply attached to the outbound request via the OAICompat `extra_headers` carrier.
- LongCat's gateway-side handling of these headers is upstream behavior, not a change requested here.
- The pre-existing test breakages in moonshot, xai, etc. (caused by the 0.9.x → 0.10.x `dify_plugin` bump) are tracked separately in the sibling PRs and are not in scope for this PR.

### DIFF
--- a/models/longcat/manifest.yaml
+++ b/models/longcat/manifest.yaml
@@ -34,4 +34,4 @@ resource:
     tool:
       enabled: true
 type: plugin
-version: 1.0.0
+version: 1.0.1
--- a/models/longcat/models/llm/_metadata.py
+++ b/models/longcat/models/llm/_metadata.py
@@ -1,0 +1,125 @@
+"""Helpers for attaching Dify metadata to LongCat requests as request headers.
+
+LongCat's plugin extends ``dify_plugin.OAICompatLargeLanguageModel``,
+whose ``_generate`` method reads ``credentials['extra_headers']`` and
+merges the entries into the outbound HTTP request headers. The Dify
+app_id is therefore propagated by writing the dify headers into that
+credential.
+
+Header values are constrained to visible ASCII (0x20-0x7E inclusive) so
+``urllib3`` never rejects the request with ``InvalidHeader`` on CR/LF
+or other control characters. Values are truncated to 256 characters
+as a hard cap (UUID app_ids are 36 chars, so the cap is a safety net
+rather than a real bound).
+
+The helper is opt-in: when ``credentials['enable_request_metadata']`` is
+not the literal string ``"enabled"``, the helper does nothing and the
+request shape is unchanged. When enabled, the helper mutates
+``credentials['extra_headers']`` in place, preserving any
+caller-supplied entries (Dify keys are written alongside, with Dify
+keys winning on collision so the helper always controls its own
+observability markers).
+
+Caveats:
+- LongCat's public API does not document these headers today, so the
+  value is purely for observability / proxy-side propagation, identical
+  to the bury-point headers used in the other LLM plugins (anthropic,
+  gemini, stepfun, openai_api_compatible, tongyi, volcengine, zhipuai,
+  minimax, cohere, mistralai, deepseek, fireworks, ollama,
+  togetherai, moonshot, groq, xai, openrouter, sagemaker, yi, mimo).
+- The helper never raises; if building the merged headers fails for
+  any reason, the call falls back to a no-op so user requests are not
+  blocked.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+_ENABLED = "enabled"
+_APP_ID_HEADER = "X-Dify-App-Id"
+_SOURCE_HEADER = "X-Dify-Source"
+_SOURCE_VALUE = "dify"
+_MAX_VALUE_LENGTH = 256
+
+
+def _normalize_header_value(s: Any) -> str:
+    """Normalize a value into a header-safe string.
+
+    Coerces non-string input via ``str()`` so that, e.g., a numeric ``0``
+    becomes ``"0"`` rather than being silently dropped by the empty-check,
+    strips CR/LF and other control characters (``urllib3`` rejects them
+    with ``InvalidHeader``), and truncates to 256 characters as a hard
+    cap.
+    """
+    if s is None:
+        return ""
+    if not isinstance(s, str):
+        s = str(s)
+    if not s:
+        return ""
+    # Strip CR/LF and other control characters; keep visible ASCII only.
+    s = "".join(ch for ch in s if 0x20 <= ord(ch) <= 0x7E)
+    if not s:
+        return ""
+    return s[:_MAX_VALUE_LENGTH]
+
+
+def build_dify_headers(app_id: Any) -> dict[str, str]:
+    """Build the Dify header dict, or return an empty dict.
+
+    Returns an empty dict if the resolved ``app_id`` is missing or
+    normalizes to the empty string, so the caller can skip attaching
+    metadata entirely.
+    """
+    app_id_normalized = _normalize_header_value(app_id)
+    if not app_id_normalized:
+        return {}
+    return {
+        _APP_ID_HEADER: app_id_normalized,
+        _SOURCE_HEADER: _SOURCE_VALUE,
+    }
+
+
+def apply_dify_headers_if_enabled(credentials: dict) -> None:
+    """Attach Dify observability headers to ``credentials['extra_headers']`` in place.
+
+    Reads ``credentials['enable_request_metadata']``; when ``"enabled"`` and
+    ``credentials['app_id']`` resolves to a non-empty string, writes the
+    Dify ``X-Dify-App-Id`` / ``X-Dify-Source`` headers into
+    ``credentials['extra_headers']``. Existing entries on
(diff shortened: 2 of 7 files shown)


Mairuis (changes requested):
At the reviewed commit, apply_dify_headers_if_enabled reads app_id from credentials. In a local helper test, enable_request_metadata=enabled with no credentials["app_id"] produced no Dify headers. I have not verified the complete deployed Dify -> daemon -> plugin invocation path, so this alone does not establish a runtime failure.

Could you confirm how app_id reaches credentials in the intended deployment? If it is available only through the current SDK session, the helper would need to read that source. A regression test that starts from the actual invocation context, rather than manually inserting app_id into credentials, would help verify the intended behavior.

This clarification replaces the earlier, overly categorical claim that normal Dify calls necessarily fail to send the metadata.

PR #3834, opened 2026-09-04
fix(deepseek): forward the user field and drop the redundant user_id (#3765)

### LINKED ISSUE #3765: missing cache hit with deepseek model 0.0.20

Taranum01:
"Fixes #3765\n\n## What problem does this PR solve?\n\nThe 0.0.20 rewrite of the DeepSeek plugin started injecting a non-standard `user_id` field into `model_parameters` while the OAICompat base class was no longer receiving the `user` argument (so its OpenAI-standard `user`-field emission was being skipped too). DeepSeek upstream caches chat-completion requests by their body, and the extra `user_id` field changed the cache key for every request \u2014 cache hit dropped from 97 % to ~75 % on long-instruction agents.\n\nPR #3805 attempted the same fix but was closed without merging. Re-attempting now against the current mainline (which has also picked up `apply_dify_headers_if_enabled` since the first attempt).\n\n## What is changed and how it works?\n\nTwo related changes, folded into the same fix:\n\n1. Stop injecting `model_parameters[\"user_id\"] = user`. The OAICompat base class already emits the OpenAI-standard `user` field on the request body, so the legacy `user_id` injection was redundant and was the only thing perturbing the upstream cache key.\n2. Forward the `user` parameter through `super()._invoke(...)`. The 0.0.20 rewrite had stopped threading the `user` argument down to the base class, so the OAICompat `user`-field emission was being skipped for every DeepSeek call. Forwarding restores the 0.0.19-era contract.\n\n## How it was tested?\n\n```\n$ uv run pytest tests/ -o addopts=\n34 passed, 8 skipped\n```\n\n1 updated test in `tests/test_llm_call.py`: `test_invoke_forwards_user_and_omits_user_id_field` (renamed from `test_invoke_uses_official_user_id_and_default_endpoint`) pins the new contract: `user` is forwarded to the base class, and `model_parameters` must NOT contain a `user_id` key (the field whose addition to the request body was perturbing the cache key).\n\n## Release Notes\n\n- fix(deepseek): stop injecting `model_parameters[\"user_id\"]` and forward the `user` argument through `super()._invoke(...)`. The 0.0.20 rewrite had (a) added a non-standard `user_id` field to the request body and (b) stopped threading the `user` argument down to the OAICompat base class, which is what emits the OpenAI-standard `user` field. (a) alone was perturbing the DeepSeek upstream cache key, dropping cache hit rate from ~97 % to ~75 % on long-instruction agents; (b) was a regression vs the 0.0.19 contract. Bumps the deepseek plugin from 0.0.21 to 0.0.22.\n"


Mairuis (approved):

Mairuis:
Please fix the CI errors. @Taranum01

Taranum01:
CI failure was a missing `## Release Notes` section in the PR body — the `Pre PR Check Per Plugin` job's `test (models/deepseek)` step runs `scripts/tools/extract-changelog.py` against the PR body and exits non-zero when the section is empty.

Added a non-empty `## Release Notes` section summarising the fix and noting the manifest bump (0.0.21 → 0.0.22). Re-running CI now; the manifest is already at 0.0.22 on this branch, so no code changes were needed.

Taranum01:
Status update: the body change (adding `## Release Notes`) should clear the failing `test (models/deepseek)` step on the next run, but `.github/workflows/pre-pr-check-per-plugin.yaml` only triggers on `pull_request_review: [submitted]` (line 3-4), so an empty commit or body edit alone won't re-run it.

Could you either:
1. Re-approve the review (which re-fires `pull_request_review: submitted` and re-runs the workflow), or
2. Click "Re-run jobs" on run 34104746746 from the Checks tab.

The code itself is unchanged from the approved state — the only thing blocking CI was the missing PR-body section, which is now in place. No source changes were made; the merge with `main` in commit `ef7a319` is already in the branch.

Taranum01 (commented):
Re-iterating that the `## Release Notes` section is now in the PR body (added ~2 hours ago) and the underlying CI failure cause is fixed. The previous failed run was triggered by the missing-body state, and the workflow doesn't re-run on body edits or empty commits. Re-submitting this comment to fire a fresh `pull_request_review: submitted` event so the workflow re-runs and confirms the fix.

Taranum01:
Found a way to re-trigger the workflow (a `pull_request_review: submitted` event fires the workflow regardless of state), but the `detect-changes` job has a guard `if: github.event.review.state == 'approved' && github.event.pull_request.base.ref == 'main' && !github.event.pull_request.draft`, so a comment-only review leaves the test job skipped and `check-complete` reports "Workflow skipped because this review does not require plugin checks" → exit 0. The new run (34220622949) `Check Complete: success` is therefore a no-op (test didn't actually execute).

To re-run the actual test on the now-correct body, the maintainer needs to:
1. Click "Re-run jobs" on run 34104746746 (or any failed run) from the Checks tab, or
2. Re-submit the approval review (which re-fires the workflow with state=approved and the test job actually runs through the Release Notes check).

Once the test job executes against the current body, the `Check Release Notes` step on line 84-93 will pass (the section is in place), and the run will be green. No code changes are needed; the merge conflict in the workflow is purely about re-running an already-fixed check.

PR #3793, opened 2026-08-31
feat: add mrscraper plugins

ai-mrscraper:
## Summary

Adds the MrScraper tool plugin for Dify.

MrScraper reliably scrapes and crawls websites, extracts public web data, and transforms it into clean, structured output for AI workflows.

The plugin provides 18 tools covering:

- Website crawling and Google SERP search
- Prompt-based, listing, and structured-data extraction
- JavaScript-rendered HTML retrieval
- Reusable scraper creation and execution
- Batch scraping
- Result and account information retrieval
- Provider-level API token validation

## Release Notes

Introduces the MrScraper tool provider for Dify with 18 tools for crawling websites, extracting structured public web data, rendering JavaScript-powered pages, creating and running reusable scrapers, and retrieving results.

MrScraper can be used in Dify Agents, Chatflows, and Workflows. A MrScraper API token is required.

## Change Type

- [ ] Documentation / non-plugin change
- [x] Non-LLM plugin (tools, extensions, datasource, etc.)
- [ ] LLM plugin

## Screenshots / Videos

Not included. Live Dify UI validation is pending.

| Before | After |
| ------ | ----- |
| MrScraper was not available | MrScraper provider with 18 tools |

## LLM Plugin Checklist

Not applicable. This is a non-LLM tool plugin.

## Version

- [x] Set the top-level `version` in `manifest.yaml` to `0.0.1`
- [ ] `dify_plugin>=0.3.0,<0.6.0` is declared in `pyproject.toml` and locked in `uv.lock`

The plugin currently pins `dify-plugin==0.10.2` and requires Dify 1.17.0 or later.

## Testing

- [x] Local deployment — Dify version: 1.17.0
- [ ] SaaS (cloud.dify.ai)

Automated validation completed:

- `uv lock --check --offline`
- `uv run pytest` — 9 tests passed
- `uv run ruff check .` — all checks passed
- Validated `manifest.yaml` and provider YAML
- Validated all 18 registered tool schemas
- Validated the bundled structured-data JSON

PR #3533, opened 2026-07-30
feat(tools): add prompt-template plugin — lightweight template render…

tanzhangjia:
…er with {{variable}} substitution

A minimal Dify plugin that renders prompt templates with workflow variables. Two input boxes (system template + user template), supports {{variable}} syntax. Outputs rendered strings.

## Change Type

- [ ] Documentation / non-plugin change
- [x] Non-LLM plugin (tools, extensions, datasource, etc.)
- [ ] LLM plugin

## Screenshots / Videos

| Before | After |
| ------ | ----- |
|        |       |

## LLM Plugin Checklist

<details>
<summary>Areas affected by this change (check all that apply)</summary>

- [ ] Message flow (system messages, user ↔ assistant turn-taking)
- [ ] Tool interaction flow (multi-round usage, Agent App and Agent Node)
- [ ] Multimodal input (images, PDFs, audio, video, etc.)
- [ ] Multimodal output (images, audio, video, etc.)
- [ ] Structured output (JSON, XML, etc.)
- [ ] Token consumption metrics
- [ ] Other LLM functionality (reasoning, grounding, prompt caching, etc.)
- [ ] New models / model parameter fixes

</details>

## Version

- [ ] Bumped top-level `version` in `manifest.yaml` (not the one under `meta`)
- [ ] `dify_plugin>=0.3.0,<0.6.0` is declared in `pyproject.toml` and locked in `uv.lock` (or kept in `requirements.txt` for legacy plugins without `uv.lock`) — [SDK docs](https://github.com/langgenius/dify-plugin-sdks/blob/main/python/README.md)

## Testing

- [x] Local deployment — Dify version: 
- [x] SaaS (cloud.dify.ai)

PR #3839, opened 2026-09-05
feat(thegrid): add The Grid model provider plugin

CastilloLuis:
## What

Adds **The Grid** as a model provider plugin under `models/thegrid`, following the `models/longcat` template.

The Grid is an inference marketplace that serves models from several labs behind one OpenAI-compatible API. Instruments address a capability tier rather than a specific lab's model name — `text-standard`, `code-prime` and `agent-max` each route to a current model for that tier — so an app keeps working when the underlying model is replaced.

It subclasses `OAICompatLargeLanguageModel` with no custom request logic: base URL, `mode`, `function_calling_type` and `stream_function_calling` only. 17 predefined instruments, `supported_model_types: [llm]`.

Donating under `author: langgenius` per this repo's convention.

## Verification

Run against the live API through the plugin class itself, not mocked:

- **All 17 predefined YAMLs parse into the SDK's `AIModelEntity`** — the same construction the runtime does at load time
- `validate_credentials("text-standard", ...)` → OK
- Streaming invoke → 72 chunks, correct content
- Tool calling on `agent-standard` → `get_weather({"city": "Paris"})`
- `_position.yaml` covers exactly the 17 model files, no drift
- `uv.lock` generated with `uv lock`

## Notes on the model definitions

Generated from `GET https://api.thegrid.ai/v1/models` rather than hand-written, so context sizes and capability flags match the API exactly. Three deliberate choices:

1. **No `pricing:` block.** The Grid is market-priced and the catalog currently returns `"pricing": null` for every instrument, so any static figure would be wrong at merge time. Left off rather than guessed.

2. **No `enable_thinking` toggle.** Reasoning is `mandatory: true` on the standard/prime tiers — I tested passing a disable flag and the instruments still emitted reasoning tokens, so a toggle would be a lie. Exposed `reasoning_effort` instead, with each instrument's actual supported values from the catalog.

3. **`kimi-latest` `max_tokens` capped at 131072.** The catalog reports `max_completion_tokens` equal to its full 1,048,576 context window, which cannot be an output ceiling; there's a comment in the file saying so.

## Endpoints

Only `/v1/chat/completions` is wired. `/v1/completions` and `/v1/embeddings` return 404 on The Grid, so `rerank`, `text_embedding`, `tts` and `speech2text` are all `false` in the manifest.

Disclosure: I work on The Grid.


Mairuis:
Thanks for the contribution and the detailed verification notes. This repository is for plugins maintained by the Dify team. Please submit The Grid as a community plugin to [langgenius/dify-plugins](https://github.com/langgenius/dify-plugins), following its [submission requirements](https://github.com/langgenius/dify-plugins/blob/main/docs/plugin-submission-requirements.md), and use your own publisher identity rather than `author: langgenius`.

Following the same routing used in #3793 and #3533, we are closing this PR here. This is a repository-scope decision, not a rejection of the integration; please open a new submission in dify-plugins.

CastilloLuis:
Understood, and thanks for the clear routing plus the pointer to the submission requirements.

Resubmitted to `langgenius/dify-plugins` as langgenius/dify-plugins#3041, under `author: the-gridai` rather than `langgenius`.

Repackaged for the community requirements rather than moved as-is: added `PRIVACY.md` (referenced from the manifest), rewrote the README with setup steps, required credentials, connection requirements and the source repository link, switched to `requirements.txt`, and dropped the lockfile so the package contains only what is needed at runtime.

PR #3864, opened 2026-09-10
fix(agent): do not promote JSON inside Thought: to tool actions

### LINKED ISSUE #3861: agent strategic bug
### Self Checks

### Dify version

the newest

### Plugin version

0.47.0

### Cloud or Self Hosted

Self Hosted (Docker)

### Steps to reproduce

[Bug] Agent strategy 0.0.47 (ReAct): a JSON snippet inside "Thought:" is misparsed as an Action → real tool call is skipped/overwritten and the truncated thought is emitted as the final answer

Models: reproduced with Qwen

Describe the bug
Since 0.0.47, the ReAct output parser promotes any JSON object that appears anywhere in the model output — including a plain "tool-input-like" JSON written inside Thought: — into an AgentScratchpadUnit.Action.

Consequences observed in one round:

The real Action: JSON of that round is overwritten by the bogus Action (the last JSON blob wins), so the intended tool is never executed;
Everything the model writes after the bogus JSON is dropped from scratchpad.thought (the thought shown in the log/answer is truncated);
When no valid Action survives, ReAct falls back to final_answer = scratchpad.thought and ends the loop, so the agent prints the (truncated) reasoning text as the final answer and the remaining iterations are never run — exactly the "thinking content is output directly and subsequent tool calls are interrupted" symptom;
When the bogus Action name is not a real tool, the round is burned on there is not a tool named <bogus> and the agent usually ends with Maximum iteration limit (n) reached.

### ✔️ Error log

Image

Image

Souravrajvi0:
## Summary

Fixes #3861.

The ReAct output parser (agent strategy v0.0.47) promoted **any** action-like JSON blob found while in `THINKING` state to a real tool call — including JSON the model wrote inside free-form `Thought:` text (e.g. when Qwen quotes an example call). That could:

- overwrite the real `Action:` JSON (last blob wins),
- truncate `scratchpad.thought` after the bogus action is set, and
- end the agent loop early with truncated reasoning as the final answer.

This change gates JSON → Action promotion so it only happens when:

1. the JSON follows an explicit `Action:` prefix, **or**
2. the model emits bare JSON before any `Thought:` marker (preserves existing support for wrapper envelopes and adjacent JSON roots).

JSON embedded in `Thought:` is kept as thought text instead.

## Test plan

- [x] `uv run pytest tests/test_cot_output_parser.py -v` (44 passed)
- [x] Added regression tests for:
  - action-like JSON inside `Thought:` with a real `Action:` line afterward
  - action-like JSON inside `Thought:` with no `Action:` line
  - real `Action:` not overwritten by later thought JSON

### DIFF
--- a/agent-strategies/cot_agent/output_parser/cot_output_parser.py
+++ b/agent-strategies/cot_agent/output_parser/cot_output_parser.py
@@ -163,24 +163,26 @@ class CotAgentOutputParser:
         json_in_string = False
         json_escape = False
         pending_action_json = False
+        json_promote_as_action = False
+        saw_thought_prefix = False
         json_stack: list[str] = []
 
         cur_state = ReactState.THINKING
         last_character = ""
 
-        def emit_json_blob(blob: str):
+        def emit_json_blob(blob: str, *, promote_as_action: bool):
             """Yield a completed JSON blob as an Action or as (flagged) text."""
             action = parse_action(blob)
-            if action is not None and cur_state is ReactState.THINKING:
+            if promote_as_action and action is not None and cur_state is ReactState.THINKING:
                 yield action
-            elif action is None and cur_state is ReactState.THINKING:
+            elif promote_as_action and action is None and cur_state is ReactState.THINKING:
                 # JSON that is not a valid action: keep it as thought text
                 # but flag the parse failure so the strategy can surface it
                 # instead of ending the round silently.
                 yield ReactChunk(cur_state, blob, parse_failed=True)
             else:
-                # In the answer state a JSON blob is part of the final answer
-                # and can never become a tool call.
+                # JSON inside Thought: (issue #3861), in the answer state, or
+                # otherwise not following an explicit Action: marker stays as text.
                 yield ReactChunk(cur_state, blob)
 
         class PrefixMatcher:
@@ -306,7 +308,7 @@ class CotAgentOutputParser:
                 # overwrite an unprocessed blob in json_cache.
                 if got_json:
                     got_json = False
-                    yield from emit_json_blob(json_cache)
+                    yield from emit_json_blob(json_cache, promote_as_action=json_promote_as_action)
                     json_cache = ""
                     in_json = False
                     json_in_string = False
@@ -333,10 +335,14 @@ class CotAgentOutputParser:
                         continue
 
                     if cur_state is not ReactState.ANSWER:
-                        yield_raw_delta, emitted_chunk, delta_consumed, _ = thought_matcher.step(delta)
+                        yield_raw_delta, emitted_chunk, delta_consumed, matched_thought_prefix = (
+                            thought_matcher.step(delta)
+                        )
                         if emitted_chunk is not None:
                             yield emitted_chunk
                         yield_delta = yield_delta or yield_raw_delta
+                        if matched_thought_prefix:
+                            saw_thought_prefix = True
                         if delta_consumed:
                             index += steps
                             continue
@@ -354,6 +360,9 @@ class CotAgentOutputParser:
                 if not in_json and delta in {"{", "["}:
                     in_json = True
                     got_json = False
+                    # Only JSON after an explicit Action: (or bare JSON at the
+                    # very start before Thought:) may become a tool call.
+                    json_promote_as_action = pending_action_json or not saw_thought_prefix
                     json_cache = delta
                     json_in_string = False
                     json_escape = False
@@ -398,7 +407,7 @@ class CotAgentOutputParser:
                 index += steps
 
         if json_cache:
-            yield from emit_json_blob(json_cache)
+            yield from emit_json_blob(json_cache, promote_as_action=json_promote_as_action)
 
         # Flush the chunk tail held back as a possible partial think tag.
         # (If the stream ended inside an unclosed think block, the buffered
--- a/agent-strategies/cot_agent/tests/test_cot_output_parser.py
+++ b/agent-strategies/cot_agent/tests/test_cot_output_parser.py
@@ -210,15 +210,55 @@ class TestReActStreamParsing(unittest.TestCase):
         actions = [r for r in results if isinstance(r, AgentScratchpadUnit.Action)]
         self.assertEqual(len(actions), 1)
 
-    def test_action_input_free_text_flags_parse_failure(self):
-        # Issue #3699 (2nd scenario): "Action_input: {...}" written as free
-        # text inside the thought instead of a properly formatted action line.
+    def test_action_input_free_text_keeps_json_in_thought(self):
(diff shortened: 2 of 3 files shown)


Graybuxiao:
Thanks for the patch — the direction is right, and gating promotion on the Action: marker is much cleaner than guessing from field names. I reproduced this gating logic on top of a pristine agent-0.0.47 and ran the bundled test suite plus a few probes. A few things worth flagging before this can land.

TL;DR

✅ It correctly stops JSON inside Thought: from becoming an Action
❌ It does not fix the symptom reported in this issue
❌ It makes the failure silent (the Action parse failed log disappears)
❌ It breaks 2 existing tests
❌ It regresses a common model pattern (canonical JSON without the Action: label)
1. The reported symptom is still there
The gate only prevents the misparse. When the model writes a plan and never emits an Action: line, the round still produces zero actions, so run_agent_state stays False (ReAct.py:163) and the loop exits; then final_answer = scratchpad.thought (ReAct.py:277) prints the plan to the user. That is exactly symptom #3 in the issue description.

Fixing this also needs a strategy-layer change: when a round yields neither an Action nor a FinalAnswer, feed a format hint back as an Observation and retry the round (run_agent_state = True) instead of terminating.

2. Two existing tests fail
code
=== test_cot_output_parser ===
FAIL: test_action_input_free_text_flags_parse_failure
      self.assertEqual(len(failed), 1)
AssertionError: 0 != 1
Ran 41 tests → FAILED (failures=1)

=== test_react ===
FAIL: test_unparseable_output_surfaces_error_log
      self.assertEqual(len(error_logs), 1)
AssertionError: 0 != 1
Ran 5 tests → FAILED (failures=1)
Root cause: non-promoted blobs fall into the else branch, which emits ReactChunk(cur_state, blob) with parse_failed defaulting to False. So round_parse_failed is never set, no ERROR log is produced, and the two tests guarding the "failures must be visible" behaviour introduced for #3699 regress.

One-line fix:

python
else:
    # not promoted: still flag unparseable JSON so the strategy can report it
    yield ReactChunk(cur_state, blob, parse_failed=action is None)
3. Functional regression: canonical JSON after Thought: gets dropped
If a model omits the Action: label but emits a perfectly well-formed payload:

code
Thought: Let me look that up.
{"action": "search", "action_input": {"q": "x"}}
it is now treated as plain text → no tool call at all, and no log either. This pattern is fairly common with weaker models and it currently works on 0.0.47. Suggest keeping a promotion path for canonical-shape payloads — e.g. decide inside emit_json_blob based on the blob content, not only at the { character.

4. The escape hatch still lets the original bug through
With json_promote_as_action = pending_action_json or not saw_thought_prefix, JSON that appears before any Thought: still goes through the unmodified, permissive parse_action:

code
{"name": "张三", "city": "北京"}
Thought: noted
Probe result: Action("张三", {}) — the original misparse is still reachable. Tightening _extract_action (for blobs that do not follow Action:, require both a recognised name key and an input key) closes this hole.

Probe summary
Input	After this patch
Thought: + canonical JSON, no Action: label	action=None, silent ❌
bare JSON, no Thought:	('search', {'q': 'x'}) ✅
Thought: + Action:	('search', {'q': 'x'}) ✅
JSON before Thought: ({"name": "张三"})	('张三', {}) ❌
screenshot shape (plan written in Thought:)	action=None, parse_failed=0 ❌

Souravrajvi0:
@Graybuxiao Thanks again for the detailed review — I pushed follow-up commits on `feat/react-thought-json-misparsing-4511` (`e834209f`) that try to cover your numbered points without re-breaking #3699 / direct thought answers.

### What changed since your comment

**Mid-line JSON inside `Thought:` (#3861)**  
Still not promoted. Line-start for the “canonical JSON after Thought” path is **newline or stream start only** (not space), so shapes like `Thought: For example {"action": ...}` stay in thought text.

**Your point 2 — silent failures / #3699**  
Non-promoted JSON that **looks like an action attempt** (action/tool/function keys or input-like keys) sets `parse_failed` again, so `round_parse_failed` and the **Action parse failed** log path can fire. Restored `test_action_input_free_text_flags_parse_failure` and `test_unparseable_output_surfaces_error_log`.  
(We did **not** use blanket `parse_failed=(action is None)` on every non-promoted blob — that would flag innocent `{"foo":"bar"}` in thought.)

**Your point 3 — canonical JSON without `Action:`**  
If the model emits well-formed action JSON on its **own line** after `Thought:`, we promote via strict `parse_action` (no legacy name fallback). Regression test: `test_canonical_json_on_new_line_after_thought_still_works`.

**Your point 4 — JSON before `Thought:`**  
Bare pre-`Thought:` blobs use stricter parsing (`reject_name_only_without_input`, no legacy name fallback), so `{"name": "张三", "city": "北京"}` is not promoted. Test: `test_json_before_thought_without_action_keys_is_not_promoted`.

**Your point 1 — strategy / symptom #3 (partial)**  
Added a **narrow** retry: if the round has no action and no streamed `FinalAnswer`, but `thought` contains **line-anchored** `"action"` / `"tool"` / `"tool_name"` JSON that did not execute, we set `Observation` to a format hint and continue the loop (`REACT_FORMAT_RETRY_HINT`). Integration test: truncated line JSON → retry → tool on next round.

`uv run pytest tests/test_cot_output_parser.py tests/test_react.py` — **52 passed** on this branch.

---

### Product decision needed (strategy layer)

I did **not** implement the broad rule “whenever a round has neither `Action` nor `FinalAnswer`, always retry with a format hint,” because that conflicts with valid behavior such as:

`Thought: The capital of France is Paris.` → today that correctly becomes the final answer with no tool call.

For **#3861 screenshot-style** output (long plan in `Thought:` only, no `Action:` / no line JSON), behavior is still mostly: loop exits and `final_answer = scratchpad.thought` unless the line-JSON retry heuristic fires.

**Can you confirm which behavior you want for maintainers?**

| Option | Behavior |
|--------|----------|
| **A** | Keep **A** as default: thought-only rounds end with thought as final answer (current + our narrow line-JSON retry). |
| **B** | **Always retry** when no `Action` and no `FinalAnswer` (your literal suggestion — acknowledge direct-answer regression unless we add more guards). |
| **C** | **Retry only on stronger signals** — e.g. line-anchored action JSON (current), and/or `round_parse_failed`, and/or extra heuristics you’d accept (keywords, max 1 retry per round, only when tools are configured, etc.). |
| **D** | Other — please specify. |

Also for **#3861**: is the primary success criterion **(1)** never mis-invoke tools from JSON inside thought (parser), **(2)** never surface a long “plan” as the user-visible final answer (strategy), or **both** with a defined priority?

Once we align on **A/B/C/D**, I can adjust `ReAct.py` in a follow-up commit on this PR. Parser-side items from your list should be addressed in the latest push; happy to run any extra probes you used and paste results.

